PR #15, opened 2018-01-23
better tests, add linting and type checking

JasonMWhite:
Please forgive the excess length of this PR, it's mostly a default `pylintrc` file with some sane defaults.

/cc @evadb @sleepycat

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,31 +44,34 @@ jobs:
       - image: circleci/python:3.6.1
       - image: mongo:3.6.2
 
-    working_directory: ~/repo
+    working_directory: ~/repo/python
 
     steps:
-      - checkout
+      - checkout:
+          path: ~/repo
 
       # Download and cache dependencies
       - restore_cache:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -44,31 +44,34 @@ jobs:
...
-          key: deps1-{{ .Branch }}-{{ checksum "python/requirements.txt" }}
+          key: deps1-{{ .Branch }}-{{ checksum "requirements.txt" }}
 
       - run:
           name: Installing dependencies
           command: |
-            python3 -m venv venv
-            . venv/bin/activate
-            pip install -r python/requirements.txt
+            python3 -m venv ~/.venv
```
buckley-w-david: Didn't notice in pull request to add the circleci config, but is there any reason for the difference in name of directory used for the virtual environment between this and geocoder? (The difference being this directory is preceeded by a ".", geocoder's is not.)

I actually prefer this to what is done in geocoder, but curious as to if there is any reason.

PR #19, opened 2018-01-24
hide connection string creator

JasonMWhite:
Improved connection string logic so we can make some module functions private. No longer needed by the test logic.


python/src/energuide/database.py
```diff
@@ -51,13 +53,18 @@ def build_connection_string(coords: DatabaseCoordinates):
     return connection_string
 
 
+@contextmanager  # type: ignore
+def mongo_client(database_coordinates: DatabaseCoordinates) -> typing.Iterable[pymongo.MongoClient]:
+    connection_string = _build_connection_string(database_coordinates)
+    yield pymongo.MongoClient(f'{connection_string}')
+
```
buckley-w-david: Am I missing something, or are you not closing the `MongoClient` connection?

JasonMWhite: Nope, not missing anything. Good call.

PR #23, opened 2018-01-25
add cli load command

JasonMWhite:
This PR adds a `load` subcommand (the first) to the `energuide` main command. It loads the supplied filename to the specified (through environment variables or command-line options) mongodb collection.

### DIFF
--- a/python/src/energuide/cli.py
+++ b/python/src/energuide/cli.py
@@ -1,2 +1,29 @@
+import click
+from energuide import database
+from energuide import transform
+
+
+@click.group()
 def main() -> None:
-    print("Hello")
+    pass
+
+
+@main.command()
+@click.option('--username', envvar=database.EnvVariables.username.value, default=database.EnvDefaults.username.value)
+@click.option('--password', envvar=database.EnvVariables.password.value, default=database.EnvDefaults.password.value)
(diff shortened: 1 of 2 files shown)


python/src/energuide/cli.py
```diff
@@ -1,2 +1,29 @@
...
+@main.command()
+@click.option('--username', envvar=database.EnvVariables.username.value, default=database.EnvDefaults.username.value)
+@click.option('--password', envvar=database.EnvVariables.password.value, default=database.EnvDefaults.password.value)
+@click.option('--host', envvar=database.EnvVariables.host.value, default=database.EnvDefaults.host.value)
+@click.option('--port', envvar=database.EnvVariables.port.value, default=database.EnvDefaults.port.value, type=int)
+@click.option('--db_name', envvar=database.EnvVariables.database.value, default=database.EnvDefaults.database.value)
+@click.option('--collection',
+              envvar=database.EnvVariables.collection.value,
+              default=database.EnvDefaults.collection.value)
+@click.option('--filename', type=click.Path(exists=True), required=True)
```
buckley-w-david: Any particular reason you've put all the authentication arguments onto the command as opposed to the group? I ask because this is different than the approach taken with geocoder.

JasonMWhite: Mainly because it's the only subcommand right now, and I'm not sure if future ones will need them. Maybe? Punt until we know.

PR #25, opened 2018-01-26
add Dwelling class

JasonMWhite:
Adds a new `Dwelling` class to be used by the transform logic. Right now it only extracts a single field, but we will add on to this class as we go. Also, this will serve as a great logical interface point for loading an iterable of `Dwelling`s to MongoDB.

### DIFF
--- a/python/src/energuide/dwelling.py
+++ b/python/src/energuide/dwelling.py
@@ -18,7 +18,7 @@ class Dwelling:
         self._house_id = house_id
 
     @staticmethod
-    def verify_schema(data: DwellingData):
+    def _verify_schema(data: DwellingData):
         for row in data:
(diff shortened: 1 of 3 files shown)


python/src/energuide/dwelling.py
```diff
@@ -0,0 +1,38 @@
...
+DwellingData = typing.Iterable[typing.Dict[str, typing.Any]]
+
+
+class Dwelling:
+
+    def __init__(self, *, house_id: int) -> None:
+        self._house_id = house_id
+
+    @staticmethod
+    def verify_schema(data: DwellingData):
```
buckley-w-david: Since it's not needed anywhere else, we can make it private

JasonMWhite: Good catch

PR #35, opened 2018-01-29
Read file

buckley-w-david:
Generator function that reads a file, and produces an iterator of Dict[str, Any] objects

### DIFF
--- a/python/src/energuide/interpreter.py
+++ b/python/src/energuide/interpreter.py
@@ -1,3 +1,4 @@
+import csv
 import datetime
 import enum
 import typing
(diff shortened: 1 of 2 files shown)


python/src/energuide/interpreter.py
```diff
@@ -113,3 +114,16 @@ def from_row(cls, row: InputData) -> 'ParsedDwellingDataRow':
...
         )
+
+
+def read(filename: str) -> typing.Iterator[InputData]:
+    with open(filename, 'r') as csvfile:
+        reader = csv.reader(csvfile)
+        #pylint: disable=stop-iteration-return
+        columns = next(reader)
+
+        def to_dict(row) -> InputData:
```
JasonMWhite: You could just embed this dict comprehension directly in the `for` loop before and get rid of the pylint exception. Or, define the inner `to_dict` function before the `open` command` - that might also get rid of the linter error.

PR #44, opened 2018-01-29
update python readme with more detailed instructions

JasonMWhite:
We will need to begin integrating ETL output and API input assumptions very soon. To that end, I've updated the README to make installation instructions more detailed for everyone's benefit.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -2,8 +2,84 @@
 
 ### Installation
 
-Inside a virtualenv, run:
+#### Python 3.6
+
+Energuide API - ETL is tested to run on Python 3.6. To begin, you will need to have Python 3.6 installed.
+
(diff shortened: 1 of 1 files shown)


python/README.md
```diff
@@ -2,8 +2,84 @@
...
+#### Running the app
+
+The ETL application is accessed from the `energuide` CLI. Run `energuide --help` for help.
+
+There is currently a single command for energuide: 
+```
+energuide load --filename /path/to/file
+```
+
+A sample file is included for demonstration purposes at `./tests/sample.csv`
```
buckley-w-david: We should likely advise them to use randomized_energuide_data.csv instead as it more accurately reflects the real data.

PR #45, opened 2018-01-29
Rewrite transform.run

buckley-w-david:
Rewrote transform.run function to use new code path.
Fixed Evaluation to_dict to return strings instead of objects for encoding purposes
Added MODIFICATIONDATE field values to randomized_energuide_data.csv


python/src/energuide/transform.py
```diff
@@ -72,8 +72,8 @@ def run(coords: database.DatabaseCoordinates,
...
         collection: str,
         filename: str) -> None:
-    chunks = pd.read_csv(filename, chunksize=CHUNKSIZE)
 
-    for chunk in chunks:
-        cleaned = clean(chunk)
-        database.load(coords, database_name, collection, cleaned)
+    raw_data = reader.read(filename)
+    grouped = reader.grouper(raw_data, dwelling.Dwelling.GROUPING_FIELD)
+    dwellings = (dwelling.Dwelling.from_group(group) for group in grouped)
```
JasonMWhite: This approach fully materialized the iterable into a tuple - weird choice. At a minimum, I'd be using a list comprehension for this.

Also, I'd rather not materialize the entire iterable in memory between the intepretation and the load phase. I'd much rather have the `grouped_load` command consume the data from the upstream iterator on demand.

JasonMWhite: I retract my comment: https://www.python.org/dev/peps/pep-0289/. Schooled.

TIL - I always thought this was a tuple comprehension, instead of a generator expression.

PR #47, opened 2018-01-29
remove now unused functions and tests

buckley-w-david:
Remove unused functions from transform.py and database.py and unnecessary tests from test_transform.py and test_database.py

remove pandas dependency from project


JasonMWhite:
Can we remove `pandas` now from the requirements file?

PR #52, opened 2018-01-30
Extract

buckley-w-david:
Create extract function to load data from a data source, provide minimal validation, and strip PII fields.

### DIFF
--- a/python/src/energuide/extractor.py
+++ b/python/src/energuide/extractor.py
@@ -5,20 +5,6 @@ import cerberus
 from energuide import reader
 
 
-REQUIRED_FIELDS = [
-    'EVAL_ID',
-    'EVAL_TYPE',
-    'ENTRYBY',
-    'CLIENTPCODE',
-    'CLIENTNAME',
-    'TELEPHONE',
-    'MAIL_ADDR',
-    'MAIL_PCODE',
-    'TAXNUMBER',
-    'CLIENTADDR',
-    'RAW_XML'
-]
-
 DROP_FIELDS = ['ENTRYBY',
                'CLIENTNAME',
                'CLIENTADDR',
@@ -28,6 +14,11 @@ DROP_FIELDS = ['ENTRYBY',
                'MAIL_PCODE',
                'TAXNUMBER',
                'RAW_XML']
+
+REQUIRED_FIELDS = DROP_FIELDS + [
+    'EVAL_ID',
+    'EVAL_TYPE'
+]
 
 _SCHEMA = {field: {'type': 'string', 'required': True} for field in REQUIRED_FIELDS}
 
@@ -46,7 +37,7 @@ def validated(data: typing.Iterable[reader.InputData], validator) -> typing.Iter
         yield document
 
 
-def _read_csv(filepath: str):
+def _read_csv(filepath: str) -> typing.Iterator[reader.InputData]:
     csv.field_size_limit(sys.maxsize)
     with open(filepath, 'r') as file:
         csv_reader = csv.DictReader(file)
--- a/python/src/energuide/extractor.py
+++ b/python/src/energuide/extractor.py
@@ -46,10 +46,16 @@ def validated(data: typing.Iterable[reader.InputData], validator) -> typing.Iter
         yield document
 
 
-def extract(csv_input: typing.Iterable[str]) -> typing.Iterator[reader.InputData]:
+def _read_csv(filepath: str):
     csv.field_size_limit(sys.maxsize)
+    with open(filepath, 'r') as file:
+        csv_reader = csv.DictReader(file)
+        for row in csv_reader:
+            yield row
 
+
+def extract(filepath: str) -> typing.Iterator[reader.InputData]:
     validator = cerberus.Validator(_SCHEMA, allow_unknown=True)
-    csv_reader = csv.DictReader(csv_input)
-    for blob in validated(csv_reader, validator):
+
+    for blob in validated(_read_csv(filepath), validator):
         yield blob
--- a/python/src/energuide/extractor.py
+++ b/python/src/energuide/extractor.py
@@ -43,7 +43,6 @@ def validated(data: typing.Iterable[reader.InputData], validator) -> typing.Iter
         document['FORWARDSORTATIONAREA'] = document['CLIENTPCODE'][:3]
         for key in DROP_FIELDS:
             document.pop(key)
-        document
 
         yield document
 
--- a/python/src/energuide/extractor.py
+++ b/python/src/energuide/extractor.py
@@ -16,14 +16,15 @@ _SCHEMA = {
         'HOUSEREGION': {'type': 'string', 'required': True},
     }
 
+_fields = _SCHEMA.keys()
+
 
 def extract(infile: str, outfile: str):
     validator = cerberus.Validator(_SCHEMA, allow_unknown=True, purge_unknown=True)
-    input_data = reader.read(infile)
-
     with open(infile, 'r') as csv_input, open(outfile, 'w') as csv_output:
         reader = csv.DictReader(csv_input)
-        writer = csv.DictWriter(csv_output, fieldnames)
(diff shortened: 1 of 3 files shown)


python/src/energuide/extractor.py
```diff
@@ -0,0 +1,61 @@
...
+    'MAIL_PCODE': {'type': 'string', 'required': True},
+    'TAXNUMBER': {'type': 'string', 'required': True},
+
+    'RAW_XML': {'type': 'string', 'required': True}
+    }
+
+DROP_FIELDS = ['HOUSEREGION',
+               'WEATHERLOC',
+               'ENTRYBY',
+               'CLIENTCITY',
```
JasonMWhite: `CLIENTCITY` and `WEATHERLOC` are fine.

python/tests/test_extractor.py
```diff
@@ -0,0 +1,13 @@
+from energuide import extractor
+
+
+def test_extract(energuide_fixture: str):
+    with open(energuide_fixture, 'r') as file_in:
```
JasonMWhite: Let's test this with an explicit in-code fixture of 1 item with the minimum set of fields, instead of referring to a test file. Test files are better for integration tests.

python/src/energuide/extractor.py
```diff
@@ -0,0 +1,56 @@
...
+import cerberus
+from energuide import reader
+
+
+_SCHEMA = {
+    'EVAL_ID': {'type': 'string', 'required': True},
+    'EVAL_TYPE': {'type': 'string', 'required': True},
+
+    'HOUSEREGION': {'type': 'string', 'required': True},
+    'WEATHERLOC': {'type': 'string', 'required': True},
```
JasonMWhite: Unless I misunderstand, I don't think the above two fields are required to be in the schema. This should be the bare minimum set: the keys needed to guarantee uniqueness, and any fields we need to drop/manipulate in the extractor itself. All other fields should be accepted and passed on.

python/src/energuide/extractor.py
```diff
@@ -0,0 +1,56 @@
...
+    'ENTRYBY': {'type': 'string', 'required': True},
+    'CLIENTCITY': {'type': 'string', 'required': True},
```
JasonMWhite: Same with `CLIENTCITY`, `MAIL_CITY`, or `MAIL_REGION`.

python/src/energuide/extractor.py
buckley-w-david: You are correct, momentary lapse on my part.

python/src/energuide/extractor.py
```diff
@@ -0,0 +1,51 @@
...
+            document.pop(key)
+
+        yield document
+
+
+def extract(csv_input: typing.Iterable[str]) -> typing.Iterator[reader.InputData]:
+    csv.field_size_limit(sys.maxsize)
+
+    validator = cerberus.Validator(_SCHEMA, allow_unknown=True)
+    csv_reader = csv.DictReader(csv_input)
```
JasonMWhite: `csv_input` is declared as just an `Iterable[str]`, but clearly it's meant to be more than that since `DictReader` expects the first `str` to be a header row, etc. You couldn't pass just any iterable of `str` in here and expect it to work, so the typing is insufficient here. (Also note that `str` is an iterable of `str` itself, so you could literally pass `foo` into this function and typing would pass).

I think we're slicing responsibility for this function at the wrong point. Maybe separate it into a separate `_read_csv` function that takes a filepath and returns an `Iterable[Dict]`?

JasonMWhite: It's OK for now to assume the input is a filepath to the whole thing. We'll work out how to switch between input fixtures and a DB connection at some point later.

python/tests/test_extractor.py
```diff
@@ -0,0 +1,31 @@
+from io import StringIO
+import pytest
+from energuide import extractor
+
+
+@pytest.fixture
+def csv_string():
+    return '''EVAL_ID,EVAL_TYPE,HOUSEREGION,WEATHERLOC,ENTRYBY,CLIENTCITY,CLIENTPCODE,CLIENTNAME,TELEPHONE,\
```
JasonMWhite: If the input to the whole thing is a filepath, let's make the fixture a filepath. You can use `tmpdir` to give you a temporary filesystem per-test for writing the fixture data to.

python/tests/test_extractor.py
```diff
@@ -0,0 +1,31 @@
...
+MAIL_CITY,MAIL_REGION,MAIL_PCODE,TAXNUMBER,RAW_XML\n123,D,Ontario,somewhere,Fred Johnson,Toronto,M5E 1W5,John \
+Fredson,999 999 9999,Toronto,Ontario,M5E 1W5,999999999999,<tag>thing</tag>'''
+
+
+def test_extract(csv_string: str):
```
JasonMWhite: This test should test three cases: the minimum set of fields, the minimum set + some other fields (`foo`, `bar`, etc), and a set of fields that does not include the minimum set.

You can use a parameterized fixture to test both of the first two cases pass.

JasonMWhite: By "this test" I mean "this test module". Probably two separate test functions.

python/src/energuide/extractor.py
```diff
@@ -0,0 +1,61 @@
...
+
+        document = validator.document
+        document['FORWARDSORTATIONAREA'] = document['CLIENTPCODE'][:3]
+        for key in DROP_FIELDS:
+            document.pop(key)
+
+        yield document
+
+
+def _read_csv(filepath: str):
```
JasonMWhite: return type annotation

python/tests/test_extractor.py
```diff
@@ -0,0 +1,69 @@
...
+Fredson,999 999 9999,123 Main st.,M5E 1W5,999999999999,<tag>thing</tag>''',
+    '''EVAL_ID,EVAL_TYPE,ENTRYBY,CLIENTADDR,CLIENTPCODE,CLIENTNAME,TELEPHONE,MAIL_ADDR,\
+MAIL_PCODE,TAXNUMBER,RAW_XML,other_1,other_2\n123,D,Fred Johnson,123 Main st.,M5E 1W5,John \
+Fredson,999 999 9999,123 Main st.,M5E 1W5,999999999999,<tag>thing</tag>,foo,bar'''])
+def passing_str(request: _pytest.fixtures.SubRequest) -> str:
+    return request.param
+
+
+@pytest.fixture
+def valid_filepath(tmpdir, passing_str: str):
```
JasonMWhite: `tmpdir` input type annotation

python/tests/test_extractor.py
```diff
@@ -0,0 +1,69 @@
...
+
+    assert 'ENTRYBY' not in item
+    assert 'CLIENTPCODE' not in item
+    assert 'CLIENTNAME' not in item
+    assert 'TELEPHONE' not in item
+    assert 'MAIL_ADDR' not in item
+    assert 'MAIL_PCODE' not in item
+    assert 'TAXNUMBER' not in item
+    assert 'CLIENTADDR' not in item
+    assert 'RAW_XML' not in item
```
JasonMWhite: Probably overkill. This will encourage the next developer to also include every new element that needs to be excluded, when that's not the purpose of the test. Just check that one is in, and one is out, and that should be sufficient.

python/tests/test_extractor.py
```diff
@@ -0,0 +1,70 @@
...
+def test_extract_missing(invalid_filepath: str) -> None:
+    with pytest.raises(reader.InvalidInputDataException) as ex:
+        output = extractor.extract(invalid_filepath)
+        _ = dict(next(output))
+
+    assert 'EVAL_TYPE' not in ex.exconly()
+    assert 'EVAL_ID' not in ex.exconly()
+
+    assert 'ENTRYBY' in ex.exconly()
+    assert 'CLIENTPCODE' in ex.exconly()
```
JasonMWhite: Same as above.

PR #57, opened 2018-01-31
Refactor extractor to output a zip file

buckley-w-david:
Change the `write_data` function of extractor to output a zip of json files instead of outputting one json file with many blobs

### DIFF
--- a/python/src/energuide/extractor.py
+++ b/python/src/energuide/extractor.py
@@ -3,6 +3,7 @@ import itertools
 import json
 import typing
 import sys
+import zipfile
 import cerberus
 from energuide import reader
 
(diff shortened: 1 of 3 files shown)


python/src/energuide/extractor.py
```diff
@@ -79,6 +81,10 @@ def __iter__(self):
...
-    output_data = _SerializableGenerator(data)
-    with open(output_path, 'w') as output_file:
-        json.dump(output_data, output_file)
+    with zipfile.ZipFile(output_path, mode='w') as output_zip:
+        for blob in data:
+            blob_id = blob.get('BUILDER')
+            if blob_id:
+                output_zip.writestr(blob_id, json.dumps(blob))
+            # else:
+                # log(f'Invalid data: {blob}, must contain field \'BUILDER\'')
```
JasonMWhite: 🔥  ?

buckley-w-david: Mostly just there to remind me/us that once logging is in place, this is somewhere that will immediately take advantage of it.

If you prefer not to include things like that, I have no problems with removing it.

PR #59, opened 2018-01-31
Zip transform

buckley-w-david:
Switch transform and load paths to use new extracted zip data instead of input csv.
Removed postal code from ParsedDwellingDataRow and Dwelling data structures as that information now gets stripped away at extraction time.


python/tests/conftest.py
```diff
@@ -2,9 +2,10 @@
 import random
 import typing
 
+import py
 import pymongo
 import pytest
-from energuide import database
+from energuide import database, extractor
```
JasonMWhite: Please put this on a separate line. Helps reduce merge conflicts and keeps diffs more focused.

PR #61, opened 2018-01-31
pull ceilings data out of raw_xml in extractor

JasonMWhite:
This PR adds the first XML parsing to the extractor. It pulls out all the `ceilings` data and adds it to each row. The data is a list of dicts containing all the needed information from each ceiling.

### DIFF
--- a/python/src/energuide/extractor.py
+++ b/python/src/energuide/extractor.py
@@ -52,7 +52,7 @@ def _extract_snippets(data: typing.Iterable[reader.InputData]) -> typing.Iterato
         doc = ElementTree.fromstring(row['RAW_XML'])
         house = doc.find('House')
         if house:
-            extra_data = snippets.snip_house(house) if house else None
+            extra_data = snippets.snip_house(house)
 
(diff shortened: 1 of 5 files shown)


python/src/energuide/extractor.py
```diff
@@ -46,6 +48,16 @@ def _read_csv(filepath: str) -> typing.Iterator[reader.InputData]:
 def _extract_snippets(data: typing.Iterable[reader.InputData]) -> typing.Iterator[reader.InputData]:
     for row in data:
         row['ForwardSortationArea'] = row['CLIENTPCODE'][:3]
+
+        doc = ElementTree.fromstring(row['RAW_XML'])
+        house = doc.find('House')
+        if house:
+            extra_data = snippets.snip_house(house) if house else None
```
buckley-w-david: `if house else None` seems unnecessary here since in the line directly above it `house` is being tested in the if statement

JasonMWhite: good catch

PR #68, opened 2018-02-01
Load ceiling

buckley-w-david:
This pull request adds functionality to transform and load ceiling data extracted in the extract component into mongodb


python/src/energuide/dwelling.py
```diff
@@ -65,6 +65,56 @@ def from_data(cls, data: str) -> 'Region':
...
+
+        )
+
+    def to_dict(self):
+        return {
+            'label': self.label,
+            'typeEnglish': self.type_english,
+            'typeFrench': self.type_french,
+            'nominalRSI': self.nominal_rsi,
+            'nominalR': self.nominal_rsi*self._RSI_RATIO if self.nominal_rsi is not None else None,
```
JasonMWhite: spaces around `*`

JasonMWhite: and parentheses if necessary

python/src/energuide/dwelling.py
```diff
@@ -65,6 +65,56 @@ def from_data(cls, data: str) -> 'Region':
...
+
+    area_metres: typing.Optional[float]
+    length_metres: typing.Optional[float]
+
+
+class Ceiling(_Ceiling):
+
+    _RSI_RATIO = 5.678263337
+    _FEET_SQUARED_RATIO = 10.7639
+    _FEET_RATIO = 3.28084
```
JasonMWhite: these aren't really ratios. More like multipliers, or conversion factors.

python/src/energuide/dwelling.py
```diff
@@ -89,25 +140,46 @@ class ParsedDwellingDataRow(_ParsedDwellingDataRow):
...
+                }
+            }
+        }
     }
 
     @classmethod
     def from_row(cls, row: reader.InputData) -> 'ParsedDwellingDataRow':
         validator = cerberus.Validator(cls._SCHEMA, allow_unknown=True)
         if not validator.validate(row):
+            # import pdb; pdb.set_trace()
```
JasonMWhite: 🔥

python/src/energuide/dwelling.py
```diff
@@ -89,25 +140,46 @@ class ParsedDwellingDataRow(_ParsedDwellingDataRow):
...
-            modification_date=parser.parse(row['MODIFICATIONDATE']),
-            year_built=row['YEARBUILT'],
-            city=row['CLIENTCITY'],
-            region=Region.from_data(row['HOUSEREGION']),
-            forward_sortation_area=row['forwardSortationArea']
+            eval_id=parsed['EVAL_ID'],
+            eval_type=EvaluationType.from_code(parsed['EVAL_TYPE']),
+            entry_date=parser.parse(parsed['ENTRYDATE']).date(),
+            creation_date=parser.parse(parsed['CREATIONDATE']),
+            modification_date=parser.parse(parsed['MODIFICATIONDATE']),
```
JasonMWhite: Should we put `parser.parse` in the `coerce` method of the validator?

python/src/energuide/dwelling.py
```diff
@@ -117,12 +189,14 @@ def __init__(self, *,
                  evaluation_type: EvaluationType,
                  entry_date: datetime.date,
                  creation_date: datetime.datetime,
-                 modification_date: datetime.datetime
+                 modification_date: datetime.datetime,
+                 ceilings: typing.List[Ceiling]
```
JasonMWhite: add a `,` at the end to minimize line edits the next time we add something

python/tests/test_dwelling.py
```diff
@@ -10,30 +10,52 @@
...
+            {
+                'label': 'Main attic',
+                'type_english': 'Attic/gable',
+                'type_french': 'Combles/pignon',
+                'nominal_rsi': '2.864',
+                'effective_rsi': '2.9463',
+                'area': '46.4515',
+                'length': '23.875',
+            }
+        ]
```
JasonMWhite: We should have `sample_input_e` start from `sample_input_d` and `merge` in a dict with the necessary changes.

python/tests/test_dwelling.py
```diff
@@ -90,6 +112,25 @@ def test_from_code(self):
     def test_from_unknown_code(self):
         assert dwelling.Region.from_data('CA') == dwelling.Region.UNKNOWN
 
+class TestCeiling:
+
+    @pytest.fixture
+    def sample(self):
```
JasonMWhite: This same example is used in `sample_input_d` as well. Define in one place and use in both?

python/src/energuide/dwelling.py
buckley-w-david: Not unless we want to make a custom validator type. The coercion is applied before the validation, so if it coerced it into a type that isn't supported by default, we'd have to write a custom validation type for it, except if it didn't throw an exception when coercing, we automatically know it's valid, so it's a weird situation

python/src/energuide/dwelling.py
```diff
@@ -65,6 +65,56 @@ def from_data(cls, data: str) -> 'Region':
...
 
+class _Ceiling(typing.NamedTuple):
+    label: typing.Optional[str]
+    type_english: typing.Optional[str]
+    type_french: typing.Optional[str]
+
+    nominal_rsi: typing.Optional[float]
+
+    effective_rsi: typing.Optional[float]
+
```
JasonMWhite: 🔥  extra linebreaks

python/tests/test_dwelling.py
```diff
@@ -6,35 +6,40 @@
...
-        'ENTRYDATE': '2018-02-01',
-        'CREATIONDATE': '2018-02-08 09:00:00',
-        'MODIFICATIONDATE': '2018-06-01 09:00:00',
-        'CLIENTCITY': 'Montreal',
-        'forwardSortationArea': 'G1A',
-        'HOUSEREGION': 'Quebec',
-        'YEARBUILT': 2001,
-    }
+def sample_input_e(sample_input_d: reader.InputData) -> reader.InputData:
+    sample_input_d.update(EVAL_TYPE='E')
```
JasonMWhite: this mutates the object `sample_input_d` which is potentially shared amongst other test fixtures for the same test. This is to be avoided.
```
output = sample_input_d.copy()
output['EVAL_TYPE'] = 'E'
return output
```

python/tests/test_dwelling.py
```diff
@@ -105,6 +126,19 @@ def test_from_row(self, sample_input_d: reader.InputData) -> None:
...
             forward_sortation_area='K1P',
+            ceilings=[
+                dwelling.Ceiling(
+                    label='Main attic',
+                    type_english='Attic/gable',
+                    type_french='Combles/pignon',
+                    nominal_rsi=2.864,
+
+                    effective_rsi=2.9463,
+
```
JasonMWhite: 🔥

PR #73, opened 2018-02-01
Refactor dwelling.py

buckley-w-david:
This pull request is to refactor dwelling to move datatypes for xml extracted data to extracted_datatypes.py

While it may seem like a large pull request, the vast majority is simply copy/paste

### DIFF
--- a/python/src/energuide/dwelling.py
+++ b/python/src/energuide/dwelling.py
@@ -4,6 +4,7 @@ import typing
 from dateutil import parser
 import cerberus
 from energuide import reader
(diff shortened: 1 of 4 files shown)


python/src/energuide/dwelling.py
```diff
@@ -4,6 +4,7 @@
 from dateutil import parser
 import cerberus
 from energuide import reader
+from energuide.extracted_datatypes import Floor, Ceiling
```
JasonMWhite: please break this up into two lines

PR #75, opened 2018-02-02
extract wall details

JasonMWhite:
This PR requires changing the base to master before merging.

This PR adds the ability to extract wall details from the house components. Combined with the new wall codes information, this provides all the wall information present in the HOIS.

### DIFF
--- a/python/pylintrc
+++ b/python/pylintrc
@@ -183,10 +183,10 @@ name-group=
 include-naming-hint=no
 
 # Regular expression matching correct function names
-function-rgx=[a-z_][a-z0-9_]{2,30}$
+function-rgx=[a-z_][a-z0-9_]{2,120}$
 
 # Naming hint for function names
-function-name-hint=[a-z_][a-z0-9_]{2,30}$
+function-name-hint=[a-z_][a-z0-9_]{2,120}$
 
 # Regular expression matching correct variable names
 variable-rgx=[a-z_][a-z0-9_]{2,30}$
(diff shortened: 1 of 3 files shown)


python/pylintrc
```diff
@@ -183,10 +183,10 @@ name-group=
 include-naming-hint=no
 
 # Regular expression matching correct function names
-function-rgx=[a-z_][a-z0-9_]{2,30}$
+function-rgx=[a-z_][a-z0-9_]{2,120}$
```
JasonMWhite: Some of the test methods easily exceed 30 characters, particularly if you name them verbosely. I don't see a good reason why test methods should have long names, so increasing this maximum to the max line width.

python/tests/test_snippets.py
```diff
@@ -54,6 +54,50 @@ def test_floor_snippet(house: etree.ElementTree) -> None:
...
+    }
+
+
+def test_user_specified_wall_snippet() -> None:
+    xml_text = """
+<House><Components>
+    <Wall>
+        <Label>Test Floor</Label>
+        <Construction>
+            <Type rValue="2.6892" nominalInsulation="3.3615">User specified</Type>
```
JasonMWhite: Note the lack of an `idref` attribute here. This was unexpected, but is because this particular feature was a "user specified" wall code.

buckley-w-david: I notice that this wall also lacks a `Measurements` node, and so has no `height` and `perimeter`, does that go along with a "user specified" wall code, or is this just testing that the snippet extractor can handle the missing node?

JasonMWhite: Kind of an extra benefit. The main purpose of the test is to check the case where there is no `idref` attribute because the value is `User specified`.

PR #79, opened 2018-02-02
[WIP] Load walls

buckley-w-david:
This pull request adds functionality to Dwelling to transform and load extracted wall data.

### DIFF
--- a/python/src/energuide/extracted_datatypes.py
+++ b/python/src/energuide/extracted_datatypes.py
@@ -29,9 +29,73 @@ class _Wall(typing.NamedTuple):
     perimeter: typing.Optional[float]
     height: typing.Optional[float]
 
+
+class _WallCode(typing.NamedTuple):
+    id: str
+    label: typing.Optional[str]
+    structure_type_english: typing.Optional[str]
+    structure_type_french: typing.Optional[str]
+    component_type_size_english: typing.Optional[str]
+    component_type_size_french: typing.Optional[str]
+
+
 _RSI_MULTIPLIER = 5.678263337
 _FEET_MULTIPLIER = 3.28084
 _FEET_SQUARED_MULTIPLIER = _FEET_MULTIPLIER**2
+
+
+class WallCode(_WallCode):
+
+    SCHEMA = {
+        'type': 'list',
+        'required': True,
+        'schema': {
+            'type': 'dict',
+            'schema': {
+                'id': {'type': 'string', 'required': True},
+                'label': {'type': 'string', 'required': True, 'nullable': True},
+                'structureTypeEnglish': {'type': 'string', 'required': True, 'nullable': True},
+                'structureTypeFrench': {'type': 'string', 'required': True, 'nullable': True},
+                'componentTypeSizeEnglish': {'type': 'string', 'required': True, 'nullable': True},
+                'componentTypeSizeFrench': {'type': 'string', 'required': True, 'nullable': True},
+            }
+        }
+    }
+
+    @classmethod
+    def from_data(cls, wall_code: typing.Dict[str, typing.Optional[str]]) -> 'WallCode':
+        return WallCode(
+            id=wall_code['id'],
+            label=wall_code['label'],
+            structure_type_english=wall_code['structureTypeEnglish'],
+            structure_type_french=wall_code['structureTypeFrench'],
+            component_type_size_english=wall_code['componentTypeSizeEnglish'],
+            component_type_size_french=wall_code['componentTypeSizeFrench'],
+        )
+
+
+class _Codes(typing.NamedTuple):
+    wall: typing.Dict[str, WallCode]
+
+
+class Codes(_Codes):
+
+    SCHEMA = {
+        'type': 'dict',
+        'required': True,
+        'schema': {
+            'wall': WallCode.SCHEMA,
+        }
+    }
+
+    @classmethod
+    def from_data(cls, codes: typing.Dict[str, typing.List[typing.Dict[str, str]]]):
+        wall_code_list = (WallCode.from_data(wall_code) for wall_code in codes['wall'])
+        wall_codes = {wall_code.id: wall_code for wall_code in wall_code_list}
+
+        return Codes(
+            wall=wall_codes
+        )
 
 
 class Wall(_Wall):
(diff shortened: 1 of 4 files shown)


python/src/energuide/extracted_datatypes.py
```diff
@@ -18,11 +18,82 @@ class _Floor(typing.NamedTuple):
     length_metres: typing.Optional[float]
 
 
+class _Wall(typing.NamedTuple):
+    label: typing.Optional[str]
```
JasonMWhite: Should this be `Optional` given it's required in `SCHEMA`?

python/src/energuide/extracted_datatypes.py
```diff
@@ -18,11 +18,82 @@ class _Floor(typing.NamedTuple):
...
+
+    def to_dict(self) -> typing.Dict[str, typing.Any]:
+        return {
+            'label': self.label,
+            'structure_type_english': self.structure_type_english,
+            'structure_type_french': self.structure_type_french,
+            'component_type_size_english': self.component_type_size_english,
+            'component_type_size_french': self.component_type_size_french,
+            'nominal_rsi': self.nominal_rsi,
+            'nominal_r': self.nominal_rsi*_RSI_MULTIPLIER if self.nominal_rsi is not None else None,
```
JasonMWhite: spaces and parentheses

python/src/energuide/extracted_datatypes.py
```diff
@@ -54,7 +125,7 @@ def from_data(cls, ceiling: typing.Dict[str, typing.Any]):
 
         )
 
-    def to_dict(self):
+    def to_dict(self) -> typing.Dict[str, typing.Any]:
```
JasonMWhite: good catch

python/tests/test_dwelling.py
```diff
@@ -148,6 +183,19 @@ def test_from_row(self, sample_input_d: reader.InputData) -> None:
...
+                    label='Second level',
+                    structure_type_english='Wood frame',
+                    structure_type_french='Ossature de bois',
+                    component_type_size_english='38x89 mm (2x4 in)',
+                    component_type_size_french='38x89 (2x4)',
+                    nominal_rsi=1.432,
+                    effective_rsi=1.8016,
+                    perimeter=42.9768,
+                    height=2.4384,
+                )
```
JasonMWhite: Also needs a test for when `constructionTypeCode` is None

python/tests/test_extracted_datatypes.py
```diff
@@ -84,3 +121,25 @@ def test_from_data(self, sample):
...
+def _dict_codes(codes: typing.Dict[str, typing.List[typing.Dict[str, str]]]
+               ) -> typing.Dict[str, typing.Dict[str, typing.Dict[str, str]]]:
+
+    return {key: {structure['id']: structure for structure in value} for key, value in codes.items()}
+
+
+class TestWall:
+
+    @pytest.fixture
+    def sample(self, sample_input_d: reader.InputData):
```
JasonMWhite: return type annotation

python/tests/test_extracted_datatypes.py
```diff
@@ -84,3 +121,25 @@ def test_from_data(self, sample):
...
+        wall = sample_input_d['walls'][0]
+        wall['nominalRsi'] = float(wall['nominalRsi'])
+        wall['effectiveRsi'] = float(wall['effectiveRsi'])
+        wall['perimeter'] = float(wall['perimeter'])
+        wall['height'] = float(wall['height'])
+        return wall
+
+    def test_from_data(self, sample, codes: typing.Dict[str, typing.List[typing.Dict[str, str]]]):
```
JasonMWhite: return type annotation

python/tests/test_dwelling.py
JasonMWhite: Might be better in `test_extracted_datatypes`

JasonMWhite:
I'm surprised we don't have a test failure somewhere about the shape of the final document to be loaded to Mongo. Does that need to be added (in a separate PR?)

Couple of small comments.

python/src/energuide/extracted_datatypes.py
buckley-w-david: Something can be both `Optional` and required if we add `nullable=True` in the schema, which is something that I meant to do.

buckley-w-david:
I think we might need to add some more to `TestDwelling.test_to_dict` to cover the shape of the final document, as that's where I think best fits that check. Right now it's fairly bare, just checking a couple of things.

python/tests/test_dwelling.py
buckley-w-david: Such a test has been added to `test_extracted_datatypes`

python/src/energuide/extracted_datatypes.py
JasonMWhite: 👍

JasonMWhite: Should we make it nullable in the schema then?

python/tests/test_extracted_datatypes.py
```diff
@@ -48,39 +80,95 @@ def sample_input_d(ceiling_input: reader.InputData,
...
         output = extracted_datatypes.Floor.from_data(sample)
         assert output.label == 'Rm over garage'
         assert output.area_metres == 9.2903
+
+    def test_to_dict(self, sample: typing.Dict[str, typing.Any]) -> None:
+        output = extracted_datatypes.Floor.from_data(sample).to_dict()
+        assert output['areaMetres'] == sample['area']
+
+
+def _dict_codes(codes: typing.Dict[str, typing.List[typing.Dict[str, str]]]
```
JasonMWhite: This duplicates private functionality from `dwelling.py`. This is usually an anti-pattern. Either the private function should be made public, or (better) the tests are modified to not need it.

python/tests/test_extracted_datatypes.py
```diff
@@ -48,39 +80,95 @@ def sample_input_d(ceiling_input: reader.InputData,
...
+        output = extracted_datatypes.Wall.from_data(wall, _dict_codes(codes)['wall'])
+        assert output.label == 'Second level'
+        assert output.perimeter == 42.9768
+        assert output.component_type_size_english is None
+
+    def test_to_dict(self,
+                     sample: typing.Dict[str, typing.Any],
+                     codes: typing.Dict[str, typing.List[typing.Dict[str, str]]]) -> None:
+        output = extracted_datatypes.Wall.from_data(sample, _dict_codes(codes)['wall']).to_dict()
+        assert output['areaMetres'] == sample['perimeter']*sample['height']
```
JasonMWhite: spaces around `*`

python/tests/test_extracted_datatypes.py
```diff
@@ -48,39 +90,89 @@ def sample_input_d(ceiling_input: reader.InputData,
...
+        wall['nominalRsi'] = float(wall['nominalRsi'])
+        wall['effectiveRsi'] = float(wall['effectiveRsi'])
+        wall['perimeter'] = float(wall['perimeter'])
+        wall['height'] = float(wall['height'])
+        return wall
+
+    def test_from_data(self,
+                       sample: typing.Dict[str, typing.Any],
+                       dict_codes: typing.Dict[str, typing.Dict[str, typing.Dict[str, str]]]) -> None:
+        output = extracted_datatypes.Wall.from_data(sample, dict_codes['wall'])
```
JasonMWhite: No, this still doesn't feel right. Interface is too complicated, requiring a special dict construction that only `Wall`  knows about, but doesn't help you to build.

How about making `from_data` an instance method, instead of a class method? This would allow you to pass `codes` (raw from the XML) into the `__init__` method, where it would be transformed into an internal representation that no-one else needs to care about. It could also be re-used across all the calls to `from_data`.

buckley-w-david: That sounds like a reasonable alternative, I'll add that in

buckley-w-david:
This PR has been broken down into PRs #88 #90 and #91, and as such will be closed.

PR #81, opened 2018-02-05
python .gitignore

buckley-w-david:
Adds a .gitignore file to the python subdirectory for exclusion rules that may conflict with node files.

### DIFF
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -1,0 +1,105 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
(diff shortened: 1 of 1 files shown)


python/.gitignore
```diff
@@ -0,0 +1,105 @@
...
+
+# mkdocs documentation
+/site
+
+# mypy
+.mypy_cache/
+
+.idea/
+
+**/.DS_Store
```
JasonMWhite: These two are developer-environment specific, and therefore should be in the developer's own global git configuration.

PR #83, opened 2018-02-05
add window code extraction

buckley-w-david:
Add snippets to extract the codes that are used by Window elements


python/src/energuide/snippets.py
```diff
@@ -95,12 +95,54 @@ def _wall_code_snippet(wall_code: etree.ElementTree) -> typing.Dict[str, typing.
         'componentTypeSizeFrench': component_type_size_french,
     }
 
+def _window_code_snipped(window_code: etree.ElementTree) -> typing.Dict[str, typing.Any]:
```
JasonMWhite: other methods are called `snippet`

python/src/energuide/snippets.py
```diff
@@ -95,12 +95,54 @@ def _wall_code_snippet(wall_code: etree.ElementTree) -> typing.Dict[str, typing.
...
+        'coatingsTintsEnglish': coatings_tints_english,
+        'coatingsTintsFrench': coatings_tints_french,
+        'fillTypeEnglish': fill_type_english,
+        'fillTypeFrench': fill_type_french,
+        'spacerTypeEnglish': spacer_type_english,
+        'spacerTypeFrench': spacer_type_french,
+        'typeEnglish': type_english,
+        'typeFrench': type_french,
+        'frameMaterialEnglish': frame_material_english,
+        'frameMaterialFrench': frame_material_french,
```
JasonMWhite: Are all of these required in the HOIS? Let's make sure to start from exactly what's necessary to replicate the information contained there. We can add to it later.

python/tests/test_snippets.py
```diff
@@ -116,6 +116,67 @@ def test_code_snippet(code: etree.ElementTree) -> None:
...
+                'frameMaterialFrench': 'Bois',
+            }, {
+                'id': 'Code 13',
+                'label': '330204',
+                'glazingTypeEnglish': 'Triple/triple with 1 coat',
+                'glazingTypeFrench': 'Triple/triple, 1 couche',
+                'coatingsTintsEnglish': 'Low-E .20 (hard1)',
+                'coatingsTintsFrench': 'Faible E .20 (Dur 1)',
+                'fillTypeEnglish': '13 mm Air',
+                'fillTypeFrench': '13 mm d\'air',
```
JasonMWhite: Prefer using `""` quotes instead of escaping a `\'`

python/src/energuide/snippets.py
buckley-w-david: They're all required, the HOIS has a `TYPE` column that is a combination of all six of these fields

python/tests/test_snippets.py
```diff
@@ -116,6 +116,67 @@ def test_code_snippet(code: etree.ElementTree) -> None:
                 'structureTypeFrench': 'Ossature de bois',
                 'componentTypeSizeEnglish': '38x89 mm (2x4 in)',
                 'componentTypeSizeFrench': '38x89 (2x4)',
-            }]
+            }],
+            'window': [{
```
JasonMWhite: This test is getting a bit long. Maybe break it up into separate tests, and not necessarily testing the whole output.

python/src/energuide/snippets.py
buckley-w-david: Slip of the fingers

python/src/energuide/snippets.py
JasonMWhite: 👍

PR #88, opened 2018-02-06
add code parsing

buckley-w-david:
Factoring out the `WallCode` and `Codes`(which will eventually have more than just wall codes) classes from #79


buckley-w-david:
[Extract/Load Wall data](https://trello.com/c/dhTPKRYq/105-extract-load-wall-data)

PR #90, opened 2018-02-06
load wall data

buckley-w-david:
This PR factors out the `Wall` class and it's usage from #79 
It also modifies the test fixture data somewhat to add a wall to one of the entries that uses a code, instead of just being User Specified.


buckley-w-david:
[Extract/Load Wall data](https://trello.com/c/dhTPKRYq/105-extract-load-wall-data)

PR #92, opened 2018-02-06
extract exterior doors

JasonMWhite:
This PR adds extraction of the raw `door` data. `Door`s are always embedded within a `Wall`, which makes logical sense.

https://trello.com/c/L9320zNP/106-extract-load-door-data


JasonMWhite:

PR #93, opened 2018-02-06
Extract window

buckley-w-david:
This PR adds extraction of the raw window data. Windows are always embedded within a wall, door, ceiling, or basement element.

https://trello.com/c/edPnfM4G/125-extract-load-window-data


buckley-w-david:
[Extract/Load Window data](https://trello.com/c/edPnfM4G/125-extract-load-window-data)

PR #95, opened 2018-02-06
transform and load doors

JasonMWhite:
Transform & load doors into the database. Differing slightly from the HOIS at this point, which aggregates door information somewhat, while this leaves each door as a separate data point. Taking the same approach with Windows.


JasonMWhite:

PR #96, opened 2018-02-06
also extract and load door labels

JasonMWhite:
As we discussed IRL, also extracting and loading door labels.


JasonMWhite:

PR #98, opened 2018-02-06
Load window

buckley-w-david:
Transform & load windows into the database
Also makes a minor fix to `glazingTypes` fields in Window code

https://trello.com/c/edPnfM4G/125-extract-load-window-data


buckley-w-david:
[Extract/Load Window data](https://trello.com/c/edPnfM4G/125-extract-load-window-data)

PR #104, opened 2018-02-07
Update door and window xpath string

buckley-w-david:
This PR changes the xpath search string for door and window components to search for a node at any depth in the tree beneath the components node.
This is because there are many possible valid paths to those nodes that were not captured with the previous configuration (such as Basement->Wall->Door->Window)

### DIFF
--- a/python/tests/test_snippets.py
+++ b/python/tests/test_snippets.py
@@ -113,6 +113,76 @@ def test_user_specified_wall_snippet() -> None:
     }
 
 
+def test_deeply_embeded_components() -> None:
+    xml_text = """
+<House><Components>
+    <Wall>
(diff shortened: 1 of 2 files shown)


JasonMWhite:
Needs a test

python/tests/test_snippets.py
```diff
@@ -113,6 +113,76 @@ def test_user_specified_wall_snippet() -> None:
...
+</Components></House>
+    """
+
+    doc = etree.fromstring(xml_text)
+    output = snippets.snip_house(doc)
+    assert len(output['windows']) == 2
+    assert len(output['doors']) == 2
+
+
+
```
JasonMWhite: 🔥  extra linebreaks

PR #107, opened 2018-02-07
Lookup individual dwelling details by account only

sleepycat:
Previously we were thinking that a combination of account number/houseId
plus postal code. This seems to be simpler and just as effective.


JasonMWhite:
🎉

PR #108, opened 2018-02-07
refactor snippets to use xpath

JasonMWhite:
This PR refactors `snippets.py` to use `lxml.etree` and `xpath()` to extract information from the source document.

### DIFF
--- a/python/src/energuide/snippets.py
+++ b/python/src/energuide/snippets.py
@@ -2,119 +2,74 @@ import typing
 from lxml import etree
 
 
-def _ceiling_snippet(ceiling: etree.ElementTree) -> typing.Dict[str, typing.Any]:
-    label = ceiling.findtext('Label')
-    type_english = ceiling.findtext('Construction/Type/English')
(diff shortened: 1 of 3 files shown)


python/src/energuide/snippets.py
```diff
@@ -2,119 +2,74 @@
...
+        'height': 'Measurements/@height',
+    })
+
+
+def snip_house(house: etree.ElementTree) -> typing.Dict[str, typing.List[typing.Dict[str, typing.Optional[str]]]]:
+    ceilings = house.xpath('Components/Ceiling')
+    floors = house.xpath('Components/Floor')
+    walls = house.xpath('Components/Wall')
+    doors = house.xpath('Components/Wall/Components/Door')
+    windows = house.xpath('Components/*/Components/Window')
```
buckley-w-david: I believe that these two need to be updated to the change in the search path introduced in #104

JasonMWhite: Thank you - that was the source of the merge conflict and I didn't update them properly.

PR #109, opened 2018-02-07
extract heating_cooling raw xml

JasonMWhite:
This PR extracts the raw `heating_cooling` xml from the source, as discussed. We will parse this raw input in the transform logic to derive the necessary attributes.


JasonMWhite:
[Extract/Load Heating data](https://trello.com/c/Ei8Ku3dg/108-extract-load-heating-data)

buckley-w-david:
Looks good to me, once the merge conflicts are resolved.

PR #114, opened 2018-02-08
Extract raw ventilation xml

buckley-w-david:
This PR extracts the raw `ventilation` xml from the source. This will be parsed and interpreted in the transform layer.


buckley-w-david:
[Extract/Load Ventilation data](https://trello.com/c/b3f6B3k8/111-extract-load-ventilation-data)

PR #115, opened 2018-02-08
Extract only raw xml ceilings snippets

JasonMWhite:
This PR moves some Transform logic out of the extractor, and puts it in the Transform layer for the `ceilings` type.

### DIFF
--- a/python/mypy.ini
+++ b/python/mypy.ini
@@ -1,0 +1,4 @@
+[mypy]
+strict_optional = False
+ignore_missing_imports = True
+warn_unused_ignores = True
--- a/python/setup.py
+++ b/python/setup.py
@@ -2,7 +2,7 @@ import setuptools
 
 
 setuptools.setup(
-    name='energuide-etl',
+    name='energuide',
     version='0.0.1',
     long_description='',
     author='CDS-SNC',
--- a/python/src/energuide/element.py
+++ b/python/src/energuide/element.py
@@ -1,0 +1,20 @@
+import typing
+from lxml import etree
+
+
+class Element:
+
+    def __init__(self, node: etree._Element) -> None:
+        self.__node = node
+
+    @classmethod
+    def from_string(cls, data: str) -> 'Element':
+        output = etree.fromstring(data)
+        return cls(output)
+
+    def findtext(self, *args, **kwargs) -> typing.Optional[str]:
+        return self.__node.findtext(*args, **kwargs)
+
+    def xpath(self, *args, **kwargs) -> typing.List[typing.Any]:
+        output = self.__node.xpath(*args, **kwargs)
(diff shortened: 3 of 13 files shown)


python/tests/test_extractor.py
```diff
@@ -113,8 +113,7 @@ def test_extract_with_snippets(tmpdir: py._path.local.LocalPath) -> None:
         writer.writerow(data)
 
     output = list(extractor.extract_data(str(input_file)))
-    assert 'ceilings' in output[0]
-    assert output[0]['ceilings'][0]['label'] == 'Attic'
+    assert output[0]['ceilings']
```
buckley-w-david: Is this assertion to just check for the existence of a non falsy `ceilings` intentional, or is there a missing `== some_value`?

JasonMWhite: Checking for falsy values (`None`)

JasonMWhite:
There's something up with the type checking - the tests aren't checking anything on the energuide side. Investigating...

python/tests/test_extracted_datatypes.py
```diff
@@ -198,25 +199,33 @@ def test_to_dict(self,
...
+            <French>Combles/pignon</French>
+        </Type>
+        <CeilingType nominalInsulation='2.864' rValue='2.9463' />
+    </Construction>
+    <Measurements area='46.4515' length='23.875' />
+</Ceiling>
+        """
+        return etree.fromstring(data)
+
+    def test_from_data(self, sample: str):
```
buckley-w-david: Isn't `sample` an `etree._Element`, not `str`?

JasonMWhite: yes, good catch. That's me trying to catch the typing issues.

python/mypy.ini
```diff
@@ -0,0 +1,4 @@
+[mypy]
+strict_optional = False
```
JasonMWhite: We should enable this ASAP, but there are too many violations right now.

python/setup.py
```diff
@@ -2,7 +2,7 @@
 
 
 setuptools.setup(
-    name='energuide-etl',
+    name='energuide',
```
JasonMWhite: I thought this might be causing some troubles locally, but it turns out not. Better to do anyway I think.

python/src/energuide/element.py
```diff
@@ -0,0 +1,18 @@
+from lxml import etree
+
+
+class Element:
```
JasonMWhite: This change is to help `mypy` catch type errors on functions that take or return `_Element` values - right now it just skips them.

python/mypy.ini
buckley-w-david: I agree

PR #122, opened 2018-02-09
expose _Element.attrib

JasonMWhite:
Exposes `_Element.attrib` through `element.Element`, to allow us to replace calls like:
```
node.xpath('@foo')[0]
```
with
```
node.attrib['foo']
```


JasonMWhite:
Yes, that's a good question. It's just wrapping `lxml` and exposing some of its members for type checking purposes, so it doesn't need unit testing. If we were to add much custom code (and one could argue that `Element.xpath` is getting close to that line) then we should add some tests for it.

PR #123, opened 2018-02-09
moving window transformation logic out of extractor

JasonMWhite:
Similar to #121 and other PRs, this PR moves some transformation logic out of the window extraction into the transformer.

### DIFF
--- a/python/src/energuide/extracted_datatypes.py
+++ b/python/src/energuide/extracted_datatypes.py
@@ -42,7 +42,7 @@ class _Door(typing.NamedTuple):
 
 
 class _Window(typing.NamedTuple):
-    label: typing.Optional[str]
+    label: str
     glazing_types_english: typing.Optional[str]
     glazing_types_french: typing.Optional[str]
     coatings_tints_english: typing.Optional[str]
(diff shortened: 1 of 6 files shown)


python/src/energuide/extracted_datatypes.py
```diff
@@ -239,66 +239,54 @@ def to_dict(self) -> typing.Dict[str, typing.Optional[float]]:
...
-            height=(window['height']* 0.001) if (window['height'] is not None) else None,
-        )
+        code_id = window.xpath('Construction/Type/@idref')
+        code = window_code[code_id[0]] if code_id else None
+
+        code_data = {field: getattr(code, field) if code else None for field in cls._CODE_FIELDS}
+        code_data['label'] = window.findtext('Label')
+        code_data['rsi'] = float(window.xpath('Construction/Type/@rValue')[0])
+        code_data['width'] = float(window.xpath('Measurements/@width')[0]) * 0.001
+        code_data['height'] = float(window.xpath('Measurements/@height')[0]) * 0.001
```
buckley-w-david: I know it was in the original code, but looking at it now those `0.001` seem like magic numbers with no context why the multiplication is being done, should we make those a constant, something like `_MILLIMETRES_MULTIPLIER`?

JasonMWhite: Good call

PR #124, opened 2018-02-09
Water heating extraction

buckley-w-david:
This PR adds functionality for as much water heating extraction as we currently have the information to do.

There exists a property in the `Primary` and `Secondary` water heater nodes (`hasDrainWaterHeatRecovery`) which indicates that an additional row is to be extracted from this node. We do not have the required information to know what that node looks like until we get more sample data that has an example of it. As such that is not being extracted.

An assertion to the water_heating extraction test is to indicate this field is `"false"`, when we have data that has these extra rows it will fail the tests and we can add in the required functionality.

https://trello.com/c/pqClWxf9/110-extract-load-water-heating-data

### DIFF
--- a/python/src/energuide/snippets.py
+++ b/python/src/energuide/snippets.py
@@ -27,6 +27,7 @@ class _HouseSnippet(typing.NamedTuple):
     heated_floor_area: str
     heating_cooling: str
     ventilation: typing.List[str]
+    water_heating: typing.List[str]
 
 
 class HouseSnippet(_HouseSnippet):
@@ -41,6 +42,7 @@ class HouseSnippet(_HouseSnippet):
             'heatedFloorArea': self.heated_floor_area,
             'heating_cooling': self.heating_cooling,
             'ventilations': self.ventilation,
+            'waterHeating': self.water_heating,
         }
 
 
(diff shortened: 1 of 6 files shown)


buckley-w-david:
[Extract/Load Water Heating data](https://trello.com/c/pqClWxf9/110-extract-load-water-heating-data)

python/src/energuide/snippets.py
```diff
@@ -178,6 +180,10 @@ def snip_house(house: etree._Element) -> HouseSnippet:
     ventilation = house.xpath('Ventilation/WholeHouseVentilatorList/Hrv')
     ventilation_strings = [etree.tostring(hrv, encoding='unicode') for hrv in ventilation]
 
+    water_heating = house.xpath('Components/HotWater/*[self::Primary or self::Secondary]')
```
JasonMWhite: Look at you with the conditionals. I would have thought `[Primary or Secondary]` would have done it, but I'll trust you on this.

JasonMWhite: Better to do this, or just `'Components/HotWater'`?

python/tests/test_snippets.py
```diff
@@ -258,3 +258,15 @@ def test_ventilation_snippet(house: etree._Element) -> None:
...
+def test_water_heating(house: etree._Element) -> None:
+    output = snippets.snip_house(house)
+    assert len(output.water_heating) == 2
+    checker = validator.DwellingValidator({
+        'water_heating': {'type': 'list', 'required': True, 'schema': {'type': 'xml', 'coerce': 'parse_xml'}}
+    }, allow_unknown=True)
+    doc = {'water_heating': output.water_heating}
+    assert checker.validate(doc)
+    for water_heating in checker.document['water_heating']:
+        assert water_heating.attrib['hasDrainWaterHeatRecovery'] == 'false'
```
JasonMWhite:
```
assert all([water_heating.attrib['hasDrainWaterHeatRecovery'] == 'false' for water_heating in checker.document['water_heating']])
```

python/src/energuide/snippets.py
buckley-w-david: If we do `Components/HotWater` then it gets a little annoying since it would make WaterHeating.from_data the first (and only) extracted datatype from_data method that returns a list of objects, instead of a singleton.

Not the end of the world of course, or even that big a deal, it just felt cleaner to me to keep that interface consistent.

JasonMWhite: I'd rather keep the extractor dumb.

PR #131, opened 2018-02-09
replace more uses of _Element

JasonMWhite:
This PR replaces the last of the `lxml._Element` uses in the codebase with `element.Element` for type checking greatness.

### DIFF
--- a/python/src/energuide/snippets.py
+++ b/python/src/energuide/snippets.py
@@ -1,5 +1,5 @@
 import typing
-from lxml import etree
+from energuide import element
 
 
 class _Codes(typing.NamedTuple):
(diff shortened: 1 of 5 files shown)


python/src/energuide/snippets.py
```diff
@@ -27,7 +27,7 @@ class _HouseSnippet(typing.NamedTuple):
     heated_floor_area: str
     heating_cooling: str
     ventilation: typing.List[str]
-    water_heating: typing.List[str]
+    water_heating: str
```
JasonMWhite: After updating to `Element` I started getting errors that led me here. Is this right?

buckley-w-david: That is correct, that is odd because I thought I had already fixed that

JasonMWhite:
Ended up adding the missing tests to `test_element.py` in the end. There's really no short cut, kids.

PR #132, opened 2018-02-10
[WIP] transform primary heating

JasonMWhite:
Just a WIP about extracting the heating systems. There's a lot of unknowns in this data, we'll probably need to experiment with the Hot2000 applications to see all of the potential values in the XML.


JasonMWhite:
FYI @buckley-w-david You might be interested in this approach to the multiple languages.

PR #134, opened 2018-02-10
refactor ventilation to use Bilingual construct

JasonMWhite:
This PR refactors the `Ventilation` transformation to provide and use new layers of abstraction:
`VentilationType` enumerates all 5 allowable ventilation types as its own primary construction.
The class variable `_VENTILATION_TRANSLATIONS` provides a dictionary of all known VentilationTypes to `Bilingual` objects, each of which contains both the `English` and `French` textual representations for each type.

Where types of objects are predefined and the possible translations are known in the code, we can use this pattern. In cases where translations may need to come from the source document itself, we can go down one layer of abstraction and store the `Bilingual` object directly in the class.


JasonMWhite:
After some more thought, I was able to massively simplify the `Translation` class down to just a `NamedTuple`. That gives us the immutability that I was looking for, with a much simpler interface.

PR #136, opened 2018-02-11
Add pagination

sleepycat:
The pagination system implemented here is one based on cursors but trying to maintain enough simplicity that this remains plausibly usable for people without needing much (any?) explanation.
The return type of dwellingsInFSA has been changes to be a PaginatedResultSet, and it now accepts two optional arguments: limit and next.

### DIFF
--- a/api/src/schema/index.js
+++ b/api/src/schema/index.js
@@ -12,6 +12,14 @@ const Schema = i18n => {
       eq: String
     }
 
+    type PaginatedResultSet {
+      hasNext: Boolean
+      hasPrevious: Boolean
+      next: String
+      previous: String
+      results: [Dwelling]
+    }
+
     type Wall {
       label: String
       structureTypeEnglish: String
@@ -103,7 +111,7 @@ const Schema = i18n => {
     type Query {
       # ${i18n.t`Details for a specific dwelling`}
       evaluationsFor(account: Int!): Dwelling
-      dwellingsInFSA(filter: Filter forwardSortationArea: ForwardSortationArea!): [Dwelling]
+      dwellingsInFSA(filter: Filter forwardSortationArea: ForwardSortationArea! limit: Int, next: String): PaginatedResultSet
     }
 
     enum Field {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
   node:
     docker:
       # specify the version you desire here
-      - image: circleci/node:7.10
+      - image: circleci/node:8.9.4
       - image: mongo:3.6.2
 
       # Specify service dependencies here if necessary
@@ -132,7 +132,7 @@ jobs:
   node_integration:
     docker:
       # specify the version you desire here
-      - image: circleci/node:7.10
+      - image: circleci/node:8.9.4
       - image: mongo:3.6.2
 
       # Specify service dependencies here if necessary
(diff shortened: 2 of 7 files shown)


api/src/schema/index.js
```diff
@@ -103,7 +111,7 @@ const Schema = i18n => {
     type Query {
       # ${i18n.t`Details for a specific dwelling`}
       evaluationsFor(account: Int!): Dwelling
-      dwellingsInFSA(filter: Filter forwardSortationArea: ForwardSortationArea!): [Dwelling]
+      dwellingsInFSA(filter: Filter forwardSortationArea: ForwardSortationArea! limit: Int, next: String): PaginatedResultSet
```
pcraig3: Couple quick questions:

1. `limit: Int,` has a trailing comma but none of the others do. Is this intentional/meaningful? 
2. We don't care about having a `previous` parameter, eh?

api/src/schema/index.js
```diff
@@ -12,6 +12,14 @@ const Schema = i18n => {
       eq: String
     }
 
+    type PaginatedResultSet {
+      hasNext: Boolean
+      hasPrevious: Boolean
+      next: String
+      previous: String
+      results: [Dwelling]
+    }
```
pcraig3: Is this a convention or anything? 
There's a [JSON API spec](http://jsonapi.org/) that has some example structure like 

```json
{
  "links": {
    "self": "http://example.com/articles",
    "next": "http://example.com/articles?page[offset]=2",
    "last": "http://example.com/articles?page[offset]=10"
  },
  "data": [{
    "type": "articles",
    "id": "1",
    "attributes": {
      "title": "JSON API paints my bikeshed!"
    }
  }]
}
``` 

I am very much not in love with this suggested response structure but am wondering if you have any thoughts.

api/src/schema/index.js
sleepycat: Regarding the comma, GraphQL has what it calls [insignificant commas](http://facebook.github.io/graphql/October2016/#sec-Insignificant-Commas). I normally don't put them for that reason, but clearly I had a lapse in consistency.

There is definitely some more testing that needs to be added to fully exercise the pagination stuff. Adding a previous parameter should happen for sure... maybe that would be a good time to remove that unsightly comma.

api/src/schema/index.js
sleepycat: In terms of convention, there is the [Relay Cusor Connections Spec](https://facebook.github.io/relay/graphql/connections.htm). This was created to provide a foundation for [Relay](https://facebook.github.io/relay/) to build automated pagination on top of. Since we are not using Relay (and only a few people are in general, and most of the momentum seems to be around Apollo at the moment) implementing the Relay spec for this endpoint feels like overkill.

I ended up trying to find the simplest cursor pagination I could (something plausibly usable directly by an end user), found [this library](https://github.com/mixmaxhq/mongo-cursor-pagination) and adopted it's [return type](https://github.com/mixmaxhq/mongo-cursor-pagination#find).

Apollo seems to have it's own [thoughts](https://dev-blog.apollodata.com/tutorial-pagination-d1c3b3ee2823) about [pagination](https://dev-blog.apollodata.com/understanding-pagination-rest-graphql-and-relay-b10f835549e7)(https://www.apollographql.com/docs/react/recipes/pagination.html#cursor-pages), which might provide some nicer porcelain over the same plumbing. Let's revisit this.

api/src/schema/index.js
pcraig3: 👍 for mvp pagination

api/src/schema/index.js
pcraig3: Cool cool, just checking. Iterate wildly and all that.

PR #139, opened 2018-02-12
move WindowCode, WallCode, and Codes to codes.py

JasonMWhite:
Continuing down the path of #138, this PR moves `WindowCode`, `WallCode`, and `Codes` to their own module. I'm not entirely happy with the structure of `Codes` and think it could use some refactoring, but this PR is limited to moving the code, updating to use the new `Bilingual` class, and updating/fixing tests.


python/src/energuide/extracted_datatypes.py
```diff
@@ -105,16 +106,16 @@ class _Door(typing.NamedTuple):
...
 class _Window(typing.NamedTuple):
     label: str
-    glazing_types_english: typing.Optional[str]
-    glazing_types_french: typing.Optional[str]
-    coatings_tints_english: typing.Optional[str]
-    coatings_tints_french: typing.Optional[str]
+    glazing_type_english: typing.Optional[str]
+    glazing_type_french: typing.Optional[str]
+    coating_tint_english: typing.Optional[str]
+    coating_tint_french: typing.Optional[str]
```
buckley-w-david: I don't mind since it reads better this way, but the original had the `coatings_tints_english` had the words as plural because that's what they are in the xml
```
<CoatingsTints code="3">
    <English>Low-E .20 (hard1)</English>
    <French>Faible E .20 (Dur 1)</French>
</CoatingsTints>
```

JasonMWhite: I'll change it back then.

JasonMWhite: Actually no, I don't think I will. Let's leave it as-is.

PR #144, opened 2018-02-12
extract crawlspace nodes

buckley-w-david:
This PR extracts `Crawlspace` nodes from the xml.
These, along with `Basement` and `Slab` nodes all comprise the BASEMENT/FOUNDATION table

https://trello.com/c/MHCxw57a/107-extract-load-basement-data

### DIFF
--- a/python/src/energuide/snippets.py
+++ b/python/src/energuide/snippets.py
@@ -29,6 +29,7 @@ class _HouseSnippet(typing.NamedTuple):
     ventilation: typing.List[str]
     water_heating: typing.Optional[str]
     basements: typing.List[str]
+    crawlspaces: typing.List[str]
 
 
 class HouseSnippet(_HouseSnippet):
@@ -44,7 +45,8 @@ class HouseSnippet(_HouseSnippet):
             'heating_cooling': self.heating_cooling,
             'ventilations': self.ventilation,
             'waterHeatings': self.water_heating,
-            'basements': self.basements
+            'basements': self.basements,
+            'crawlspaces': self.crawlspaces,
         }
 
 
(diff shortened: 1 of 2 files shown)


python/src/energuide/snippets.py
```diff
@@ -80,6 +83,7 @@ def snip_house(house: element.Element) -> HouseSnippet:
         ventilation=ventilation_strings,
         water_heating=water_heating_string,
         basements=[node.to_string() for node in basements],
+        crawlspaces=[node.to_string() for node in crawlspaces],
```
JasonMWhite: Instead of adding both basements and crawlspaces to the raw data, what about a single entry for all basement-style nodes?

buckley-w-david: The choice comes down to if we consider crawlspaces and slabs to be "basement-style". They end up in the same table in the report, and they are structured similarly, but not enough that they would be parsed by the same code.

It would work fine, but it'd essentially make it so `Basement.from_data(...)` does a check for what kind of basement it is parsing, and passes the node down 1 of 3 code paths.

Does that sound better to to you than doing them separately?

JasonMWhite: I have no strong opinions on this. Go with your best judgement.

buckley-w-david: I don't either. I guess I'll just see how it looks when I get started on crawlspaces and slabs, and see if I want to integrate them into what I've done for basements.

buckley-w-david:
[Extract/Load Basement data](https://trello.com/c/MHCxw57a/107-extract-load-basement-data)

PR #149, opened 2018-02-12
Add ./bootstrap.sh file, refactor Makefiles

pcraig3:
**Note**

I'm going to add information about this in the README but at the minute I'm just pushing up the code so that it's visible.  Anything looks odd, let me know. 

---

The "get up and going" process is (should be) much easier now.

The `./bootstrap.sh` file installs all the things we need to install once (pyenv, mongodb, etc), and then the Makefiles have all the commands we need to run multiple times.

- The `/python/Makefile` runs everything without tripping over virtualenv problems now.
- The `/api/Makefile` looks for an `NRCAN_ENGINE_API_KEY` to be set or else it won't run.

Also, I've tested this on my Macbook the best that I can. In theory, it should work on everyone's computers as well, but I haven't been able to confirm that it does.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+if test ! $(which brew); then
+    echo "ERROR: It looks like you need to install Homebrew first."
+    exit 1 # terminate with error
+fi
+
+# Update homebrew recipes
+echo "----> Updating Homebrew"
+brew update
+
+# Pyenv
+if [[ ! $(brew ls pyenv --versions) ]]; then
+    echo "----> Installing pyenv with Homebrew"
+    brew install pyenv
+
+    echo ''
+    echo ''
+    echo '**IMPORTANT**'
+    echo 'Please make sure to put '"'"'eval "$(pyenv init -)"'"'"' near the bottom of your shell configuration file (eg, ~/.bash_profile or ~/.zshrc).'
+    echo 'For more information, refer to pyenv installation instructions.'
+    echo ''
+    echo ''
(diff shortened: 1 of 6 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+if test ! $(which brew); then
+    echo "ERROR: It looks like you need to install Homebrew first."
```
JasonMWhite: This means we are only compatible with MacOS from this point on, at least from an automation POV. This might be a problem for NRCan maintaining it.

pcraig3: Yeah, this will work for us but it's not very portable. 

Since it's only used to install Python 3 and Mongo, this shouldn't be a deal-breaker for us. The stuff in the Makefiles *should be* platform-agnostic.

python/Makefile
```diff
@@ -1,13 +1,27 @@
...
-	python3 -m venv env
-	source env/bin/activate
-	pip install -q -r requirements.txt
-	pip install -q -e .
-import_data:
+VIRTUALENV_ROOT := $(shell [ -z $$VIRTUAL_ENV ] && echo $$(pwd)/env || echo $$VIRTUAL_ENV)
+
+.PHONY: virtualenv
+virtualenv:
+	[ -z $$VIRTUAL_ENV ] && [ ! -d env ] && python3 -m venv env || true
```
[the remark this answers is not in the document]
JasonMWhite: Talked IRL

pcraig3:
Closing this until I work out what do do about mongo in the Makefile (ie, right now it relies on homebrew).

PR #152, opened 2018-02-13
refactor Window

sastels:
Move Window into its own module.

### DIFF
--- a/python/src/energuide/embedded/window.py
+++ b/python/src/energuide/embedded/window.py
@@ -1,0 +1,64 @@
+import typing
+from energuide.embedded import area
+from energuide.embedded import code
+from energuide.embedded import insulation
+from energuide.embedded import distance
+from energuide import element
+
+_MILLIMETRES_TO_METRES = 1000
+
+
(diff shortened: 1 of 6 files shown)


python/src/energuide/embedded/window.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def _window_area(self) -> area.Area:
+        return area.Area(self.width.metres * self.height.metres)
+
+    def to_dict(self) -> typing.Dict[str, typing.Any]:
+        return {
+            'label': self.label,
+            'rsi': self.window_insulation.rsi,
+            'rvalue': self.window_insulation.r_value,
```
buckley-w-david: What's your opinion on renaming the `rvalue` and `rsi` fields? As they are it's a bit inconsistent with the insulation names from other datatypes.

sastels: As discussed IRL, I'll push this as is and we'll do another refactor adding "insulation" here and other places later. cc @evadb @sleepycat

PR #153, opened 2018-02-13
refactor Door

evadb:
Move door to it's own module

### DIFF
--- a/python/src/energuide/embedded/door.py
+++ b/python/src/energuide/embedded/door.py
@@ -1,14 +1,15 @@
 import typing
 from energuide import bilingual
 from energuide import element
-
+from energuide.embedded import distance
+from energuide.embedded import area
 
 class _Door(typing.NamedTuple):
     label: str
     door_type: bilingual.Bilingual
     rsi: float
-    height: float
-    width: float
+    height: distance.Distance
+    width: distance.Distance
 
 
 class Door(_Door):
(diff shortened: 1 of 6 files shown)


python/src/energuide/embedded/door.py
```diff
@@ -0,0 +1,62 @@
...
+from energuide import bilingual
+from energuide import element
+
+
+class _Door(typing.NamedTuple):
+    label: str
+    door_type: bilingual.Bilingual
+    rsi: float
+    height: float
+    width: float
```
buckley-w-david: For `rsi`, `height`, and `width`, we've run into those enough times that we've factored out code in the same way as `bilingual`. 
There is `energuide.embedded.distance` and `energuide.embedded.insulation` that you should be using. Take a look at `energuide.embedded.ceiling.py` for an example

python/src/energuide/embedded/door.py
```diff
@@ -0,0 +1,62 @@
...
+    @property
+    def u_factor_imperial(self) -> float:
+        return self.u_factor / self._RSI_MULTIPLIER
+
+    @property
+    def area_metres(self) -> float:
+        return self.height * self.width
+
+    @property
+    def area_feet(self) -> float:
```
buckley-w-david: Similar comment as the previous, there is `energuide.embedded.area` that you should use that will handle the conversions and such for you.

PR #157, opened 2018-02-13
Remove unused test data

sleepycat:
Tests now run against data generated by the ETL process. This is an old
fixture that is no longer needed.


pcraig3:
Yeah, looks sweet.

PR #164, opened 2018-02-14
Fix linting rules

pcraig3:
1. Linting wasn't finding the files, so removing a `,` after `src/**,` in the package.json fixed it
2. Then it was finding some files we didn't want to lint, so I added a flag to ignore the autogenerated stuff in `locale`

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -15,7 +15,7 @@
     "extract": "lingui extract",
     "compile": "lingui compile",
     "precommit": "lint-staged",
-    "lint": "eslint src/**, test/**"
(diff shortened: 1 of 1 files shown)


pcraig3:
Originally pushed a commit that would break the linting rules to check it would get caught by the CI.

Good to go now.

## screenshots

### '==' should be caught by the linter

screen shot 2018-02-14 at 4 46 25 pm

### and then it is

screen shot 2018-02-14 at 4 46 44 pm

console output 👇

screen shot 2018-02-14 at 4 46 58 pm

PR #165, opened 2018-02-14
[DEMO-1] Create /welcome page for a "show-the-thing" demo

pcraig3:
The diff includes some prettify formatting changes to server.js, but the only functional changes are from lines 38-45 which add a pretty minimal `/welcome` page we can use for the demo.

## Screenshot

screen shot 2018-02-14 at 5 43 20 pm

### why add code just for demo purposes?

A few reasons, I think. Other suggestions were to demo a redirect rule or a change to the API output.

- demoing a redirect is less visual (ie, less impactful during a presentation)
- similarly, demoing an API response change is probably less not very interesting for a non-technical person
- we're agile, so it's easy for us to remove stuff we don't need

### DIFF
--- a/api/src/server.js
+++ b/api/src/server.js
@@ -3,10 +3,10 @@ import bodyParser from 'body-parser'
 import cors from 'cors'
 import { graphqlExpress, graphiqlExpress } from 'apollo-server-express'
 import Schema from './schema'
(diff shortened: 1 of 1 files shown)


evadb:
Please never change this title

pcraig3:
Decided to close this after chatting in real life.

PR #167, opened 2018-02-14
Add live API links to the README

pcraig3:
Not sure if the URL is pointing to the Azure instance yet, but once it does, we should put the link in the README.


pcraig3:
Closing this until I get a confirmation that we want to do this.

PR #168, opened 2018-02-14
[WIP] Logging

buckley-w-david:
This PR adds basic logging and more advanced exception generation/handling.

It is currently marked as [WIP] due to the expectation that things need to change, while not knowing exactly what those changes are.


buckley-w-david:

python/src/energuide/cli.py
```diff
@@ -27,7 +27,8 @@ def load(username: str, password: str, host: str, port: int, db_name: str, colle
         port=port
     )
 
-    transform.run(coords, db_name, collection, filename)
+    stats = transform.run(coords, db_name, collection, filename)
+    print(stats)
```
JasonMWhite: We should never be `print`ing

JasonMWhite:
Try looking at https://github.com/GoC-Spending/fuzzy-tribble/blob/a37d2d096bd281fe78480d6e2b164be55c23332f/src/tribble/cli.py#L13 for an alternative.

JasonMWhite:
I think the huge bad fixture is overkill in this case. I'd rather test it at a finer-grained unit level, and this might be a good case for a mock to test it at the CLI level.

We should throw an exception at some deep level of the parsing, say inside `ceiling.py`. That should bubble up and be caught at a higher level, say while iterating inside `dwelling.py`.

buckley-w-david:
Most/all relevant code has been extracted from the PR and merged into master, as such closing.

PR #171, opened 2018-02-15
Expose floor data in the schema

sleepycat:
As advertised, exposes newly added floor data via the schema.

### DIFF
--- a/api/src/schema/index.js
+++ b/api/src/schema/index.js
@@ -19,6 +19,18 @@ const Schema = i18n => {
       airFlowRateCfm: Float
     }
 
+    type Floor {
+      label: String
+      nominalRsi: Float
+      nominalR: Float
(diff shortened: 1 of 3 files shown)


api/src/schema/index.js
```diff
@@ -19,6 +19,18 @@ const Schema = i18n => {
       airFlowRateCfm: Float
     }
 
+    type Floor {
+      label: String
+      nominalRsi: Float
```
JasonMWhite: this should be `insulationNominalRsi`

api/src/schema/index.js
```diff
@@ -19,6 +19,18 @@ const Schema = i18n => {
...
+      nominalR: Float
+      effectiveRsi: Float
```
JasonMWhite: `insulationEffectiveRsi`

PR #173, opened 2018-02-15
centralize custom exceptions

buckley-w-david:
This PR starts peeling off sections of #168. It adds a new file that stores custom exceptions used in `energuide`

### DIFF
--- a/python/src/energuide/exceptions.py
+++ b/python/src/energuide/exceptions.py
@@ -1,0 +1,14 @@
+class InvalidGroupSizeException(Exception):
+    pass
+
+
+class InvalidInputDataException(Exception):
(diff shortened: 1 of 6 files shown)


python/src/energuide/exceptions.py
```diff
@@ -0,0 +1,14 @@
+class InvalidGroupSizeException(Exception):
+    pass
+
+
+class InvalidInputDataException(Exception):
+    pass
+
+
+class InvalidEmbeddedDataTypeException(Exception):
```
JasonMWhite: We should probably have a custom base class for all our custom exceptions, so at a root level we can just catch the one base class.

PR #179, opened 2018-02-16
Refactor ventilation

buckley-w-david:
Move ventilation into its own module.

### DIFF
--- a/python/src/energuide/extracted_datatypes.py
+++ b/python/src/energuide/extracted_datatypes.py
@@ -1,4 +1,3 @@
-import enum
 import typing
 from energuide import element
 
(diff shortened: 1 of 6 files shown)


python/src/energuide/extracted_datatypes.py
```diff
@@ -109,10 +94,8 @@ class _WaterHeating(typing.NamedTuple):
 
 
 _RSI_MULTIPLIER = 5.678263337
-_CFM_MULTIPLIER = 2.11888
 _FEET_MULTIPLIER = 3.28084
 _FEET_SQUARED_MULTIPLIER = _FEET_MULTIPLIER**2
-_MILLIMETRES_TO_METRES = 1000
```
sastels: `_MILLIMETRES_TO_METRES` was just something that should have been removed before?

buckley-w-david: Yes

PR #181, opened 2018-02-16
Update api dependencies

sleepycat:
As advertised.


pcraig3:
Works for me! Only one major version change is for linting, which is still working. 

🚢 🚢 🚢

PR #187, opened 2018-02-16
Floor exceptions

buckley-w-david:
add floor exception handling and tests


buckley-w-david:

python/tests/embedded/test_floor.py
```diff
@@ -19,6 +20,35 @@ def sample_raw() -> element.Element:
     """
     return element.Element.from_string(doc)
 
+BAD_XML_DATA = [
```
sastels: Could you add a comment saying what's bad about each XML block?

PR #190, opened 2018-02-16
Wall exceptions

buckley-w-david:
Add wall exception handling and tests


python/tests/embedded/test_wall.py
```diff
@@ -21,6 +22,34 @@ def raw_sample() -> element.Element:
     """
     return element.Element.from_string(doc)
 
+BAD_XML_DATA = [
```
sastels: Could you add a comment explaining what's wrong with each XML block?

PR #196, opened 2018-02-16
add crypt_utils

sastels:
Two utility functions used to sign the nrcan files.


buckley-w-david:
Have you heard of [`secrets`](https://docs.python.org/3/library/secrets.html) library in python 3.6? May or may not simplify the stuff you're doing.

JasonMWhite:
This feels like 2 PRs, and 👍 to re-using standard crypto libraries.

buckley-w-david:
If we end up needing to do a non-trivial amount of cryptography, we should probably take a look at external libraries (Found the [`cryptography`](https://cryptography.io/en/latest/) library mentioned in the [FAQ](https://www.python.org/dev/peps/pep-0506/#frequently-asked-questions) of the `secrets` PEP)

sastels:
@buckley-w-david Looks like `secrets` now provides the recommended way of getting secure random bytes, rather than via`os.urandom()` 👍

sastels:

extract_endpoint/src/crypt_utils.py
```diff
@@ -0,0 +1,13 @@
+import os
+import hmac
+import base64
+
+
+def get_salt(num_bytes=64) -> str:
```
JasonMWhite: Is this function still needed?

extract_endpoint/src/crypt_utils.py
```diff
@@ -0,0 +1,13 @@
...
+    return base64.b64encode(os.urandom(num_bytes)).decode('utf-8')
```
JasonMWhite: Ask because you mentioned that use of `os.urandom` was no longer necessary.

extract_endpoint/src/crypt_utils.py
```diff
@@ -0,0 +1,13 @@
+import secrets
+import hmac
+import base64
+
+
+def get_salt(num_bytes=64) -> str:
+    return base64.b64encode(secrets.token_bytes(num_bytes)).decode('utf-8')
+
+
+def sign_string(salt: str, key: str, data: str) -> str:
```
JasonMWhite: Is it important that the caller knows to generate a random salt and pass it in, or could this function encapsulate that functionality as well? Maybe making `salt` default to `None` and auto-generate a salt the right way if none is passed in?

extract_endpoint/src/crypt_utils.py
sastels: huh, that was just to convert the random bytes to a string to pass to the microservice, but we can use `secrets.token_hex()` instead of `secrets.token_bytes()` and get a string directly. 🔥

extract_endpoint/src/crypt_utils.py
sastels: I'd say it's not that important? ie leave the user the option of not having a salt. The security drops, but not all that much against most types of attack.

extract_endpoint/tests/test_crypt_utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.fixture
+def sample_data() -> str:
+    return 'sample data'
+
+
+def test_sign_string(sample_salt: str, sample_key: str, sample_data: str) -> None:
+    actual = sign_string(salt=sample_salt, key=sample_key, data=sample_data)
+    assert actual == '8a2dbe236ead75dd242d3ee015b3d8b6'
```
JasonMWhite: Nitpick, but this is a rather fragile test. Not sure what the best approach would be (assert it exists and is a string of a certain length?). Not a big deal.

extract_endpoint/src/crypt_utils.py
```diff
@@ -0,0 +1,9 @@
+import hmac
+import typing
+
+
+def sign_string(salt: typing.Optional[str], key: str, data: str) -> str:
```
JasonMWhite: if it's optional, it should default to `None`. This code would be simpler if it defaulted to `''` however, and was never `Optional[str]`

extract_endpoint/tests/test_crypt_utils.py
sastels: Yeah, that's not a bad idea. Also added tests ensuring that changing either the salts or the keys give different signatures.

extract_endpoint/src/crypt_utils.py
```diff
@@ -0,0 +1,7 @@
+import hmac
+
+
+def sign_string(data: str, key: str, salt='') -> str:
```
JasonMWhite: `salt: str = ''`

PR #197, opened 2018-02-16
Add azure utils

sastels:
Add and test code to push to Microsoft Azure.

~~Currently the test uses the real Azure (with keys in the local environment). Emulator code will be added.~~


sastels:

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,38 @@
...
+class EnvVariables(enum.Enum):
+    account = 'EXTRACT_ENDPOINT_STORAGE_ACCOUNT'
+    key = 'EXTRACT_ENDPOINT_STORAGE_KEY'
+    container = 'EXTRACT_ENDPOINT_CONTAINER'
+    domain = 'EXTRACT_ENDPOINT_STORAGE_DOMAIN'
+
+
+class DefaultVariables(enum.Enum):
+    account = 'devstoreaccount1'
+    key = 'Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=='
```
JasonMWhite: Is this a real key to Azure blob storage?

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.fixture
+def sample_block_blob_service(sample_storage_coordinates: StorageCoordinates) -> BlockBlobService:
+    account, key, _, domain = sample_storage_coordinates
+    return BlockBlobService(account_name=account, account_key=key, custom_domain=domain)
+
+
+@pytest.fixture
+def sample_stream_content()->str:
```
JasonMWhite: formatting

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    return 'Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=='
+
+
+@pytest.fixture
+def sample_container() -> str:
+    return 'energuide-extracted-data'
+
+
+@pytest.fixture
+def sample_domain() -> Optional[str]:
```
JasonMWhite: This never returns `None`, so don't mark it `Optional`

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,38 @@
+import enum
+from io import BytesIO
+from typing import Optional, NamedTuple
```
JasonMWhite: prefer `import typing`, `import io`, etc.

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,38 @@
...
+from azure.storage.blob import BlockBlobService
```
JasonMWhite: `from azure.storage import blob`

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def upload_stream_to_azure(coords: StorageCoordinates, stream: BytesIO, filename: str) -> str:
+    account, key, container, domain = coords
+    azure_path = filename
+
+    block_blob_service = BlockBlobService(account_name=account, account_key=key, custom_domain=domain)
+    block_blob_service.create_blob_from_stream(container, azure_path, stream)
+
+    # check Azure to make sure the file is there
```
JasonMWhite: :fire:

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,38 @@
...
+    if azure_path not in [blob.name for blob in block_blob_service.list_blobs(container)]:
+        return "upload failed"
+    return "upload succeeded"
```
JasonMWhite: This is a strange way to effectively return a boolean. Why not `return azure_path in [blob.name for blob in block_blob_service.list_blobs(container)]`

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,71 @@
...
+def sample_storage_coordinates(sample_account: str,
+                               sample_key: str,
+                               sample_container: str,
+                               sample_domain: Optional[str]) -> StorageCoordinates:
+    return StorageCoordinates(account=sample_account, key=sample_key, container=sample_container, domain=sample_domain)
+
+
+@pytest.fixture
+def sample_block_blob_service(sample_storage_coordinates: StorageCoordinates) -> BlockBlobService:
+    account, key, _, domain = sample_storage_coordinates
```
JasonMWhite: This is actually a bit dangerous. It will break if someone adds a new field to `StorageCoordinates`.

extract_endpoint/src/azure_utils.py
buckley-w-david: I believe (but Steve should verify), that that is the key to the local emulator.

sastels: yes, local emulator

extract_endpoint/src/azure_utils.py
sastels: 👍  on io, but I changed `import typing` to `from typing import Optional, NamedTuple` to fix a pylint bug (see https://github.com/PyCQA/pylint/issues/1418) :(

extract_endpoint/src/azure_utils.py
JasonMWhite: I'm confused. Is that the key to everyone's local emulator? Does it ever change?

buckley-w-david:
As discussed, this should also be refactored into package(s) under the `src` directory

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,73 @@
...
+@pytest.fixture
+def sample_domain() -> str:
+    return 'http://127.0.0.1:10000/devstoreaccount1'
+
+
+@pytest.fixture
+def sample_storage_coordinates(sample_account: str,
+                               sample_key: str,
+                               sample_container: str,
+                               sample_domain: typing.Optional[str]) -> azure_utils.StorageCoordinates:
```
JasonMWhite: `sample_domain` is never Optional

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_upload_stream_to_azure(sample_storage_coordinates: azure_utils.StorageCoordinates,
+                                sample_block_blob_service: blob.BlockBlobService,
+                                sample_container: str,
+                                sample_stream: io.BytesIO,
+                                sample_stream_content: str,
+                                sample_filename: str) -> None:
+    if sample_filename in [blob.name for blob in sample_block_blob_service.list_blobs(sample_container)]:
+        sample_block_blob_service.delete_blob(sample_container, sample_filename)
```
JasonMWhite: This looks like it's doing teardown of the previous test run. This is a test code smell. This test should be given a test fixture that it can rely on being completely clean.

extract_endpoint/src/azure_utils.py
sastels: The emulator is preconfigured to use this for the key (along with the account name `devstoreaccount1`)

JasonMWhite: > The emulator has only one built in account that you cannot edit, the account details are as follows:
Account name: devstoreaccount1
Account key: 
Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==

http://blog.podrezo.com/azure-storage-emulator/
👍

extract_endpoint/tests/test_azure_utils.py
sastels: yeah, that makes things cleaner 👍 . I moved the code to ensure a clean database up into the `sample_block_blob_service` fixture.

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,35 @@
...
+class DefaultVariables(enum.Enum):
+    account = 'devstoreaccount1'
+    key = 'Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=='
+    container = 'endpoint'
+    domain = None
+
+
+def upload_stream_to_azure(coords: StorageCoordinates, stream: io.BytesIO, filename: str) -> bool:
+    account, key, container, domain = coords
+    azure_path = filename
```
JasonMWhite: This statement seems pointless?

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+    block_blob_service = blob.BlockBlobService(account_name=account, account_key=key, custom_domain=domain)
```
JasonMWhite: if you use `coords.account`, `coords.key`, and `coords.domain` it would be less fragile.

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,35 @@
...
+import io
+import typing
+from azure.storage import blob
+
+
+class StorageCoordinates(typing.NamedTuple):
+    account: str
+    key: str
+    container: str
+    domain: typing.Optional[str]
```
JasonMWhite: Does this need to be optional?

sastels: Yes, if you set the domain to `None` then it connects to the real Azure.

extract_endpoint/src/azure_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class DefaultVariables(enum.Enum):
+    account = 'devstoreaccount1'
+    key = 'Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=='
+    container = 'endpoint'
+    domain = None
+
+
+def upload_stream_to_azure(coords: StorageCoordinates, stream: io.BytesIO, filename: str) -> bool:
```
buckley-w-david: `stream` doesn't have to be a BytesIO object, it can be any byte file-like object.
I recommend [typing.IO[bytes]](https://docs.python.org/3/library/typing.html#typing.io)

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+@pytest.fixture
+def sample_block_blob_service(sample_storage_coordinates: azure_utils.StorageCoordinates) -> blob.BlockBlobService:
+    block_blob_service = blob.BlockBlobService(account_name=sample_storage_coordinates.account,
+                                               account_key=sample_storage_coordinates.key,
+                                               custom_domain=sample_storage_coordinates.domain)
+    container = sample_storage_coordinates.container
+    for filename in [blob.name for blob in block_blob_service.list_blobs(container)]:
+        block_blob_service.delete_blob(container, filename)
+    return block_blob_service
```
JasonMWhite: This is better. To ensure consistency, ideally we should be sure that there are no side-effects from running a test. This deletion logic should be in a fixture teardown phase, instead of a fixture setup phase.

```
block_blob_service = blob.BlockBlobService(account_name=sample_storage_coordinates.account,
    account_key=sample_storage_coordinates.key, custom_domain=sample_storage_coordinates.domain)
yield block_blob_service

container = sample_storage_coordinates.container
for filename in [blob.name for blob in block_blob_service.list_blobs(container)]:
    block_blob_service.delete_blob(container, filename)
```

extract_endpoint/src/azure_utils.py
JasonMWhite: 👍

extract_endpoint/requirements.txt
```diff
@@ -1,4 +1,8 @@
+astroid==1.5.3
```
JasonMWhite: Ah yes, I knew I had encountered this error before (twice, actually), but couldn't remember how I had solved it. This is the right solution :)

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    return 'http://127.0.0.1:10000/devstoreaccount1'
+
+
+@pytest.fixture
+def sample_storage_coordinates(sample_account: str,
+                               sample_key: str,
+                               sample_container: str,
+                               sample_domain: str) -> azure_utils.StorageCoordinates:
+    return azure_utils.StorageCoordinates(account=sample_account, key=sample_key,
+                                          container=sample_container, domain=sample_domain)
```
JasonMWhite: Do we need the above 4 fixtures, or can we get rid of them by putting these constants directly in this line? Generally, they should only be separate fixtures if something else also needs them separately from the `StorageCoordinates` fixture.

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -0,0 +1,70 @@
...
+def sample_block_blob_service(sample_storage_coordinates: azure_utils.StorageCoordinates) -> blob.BlockBlobService:
+    block_blob_service = blob.BlockBlobService(account_name=sample_storage_coordinates.account,
+                                               account_key=sample_storage_coordinates.key,
+                                               custom_domain=sample_storage_coordinates.domain)
+    container = sample_storage_coordinates.container
+    for filename in [blob.name for blob in block_blob_service.list_blobs(container)]:
+        block_blob_service.delete_blob(container, filename)
+    return block_blob_service
+
+@pytest.fixture
```
JasonMWhite: Two blank lines between functions/classes at the root level of the module, please. I really need to write a custom lint rule to catch this.

extract_endpoint/tests/test_azure_utils.py
sastels: 👍 similar to what you did in the gitlaw code.

JasonMWhite:
Please squash commits before/when merging

sastels:
yeah @buckley-w-david just looking at that now. Think I'll merge this PR as is, then open a new PR to move crypt_util and azure_util code to the proper place, make sure testing works, etc. Thoughts?

buckley-w-david:
That sounds fine, just wanted to make sure that it wasn't forgotten

PR #199, opened 2018-02-16
Circleci for endpoint

evadb:
This PR adds a circleCI step for extract_endpoint 🎉  I think with this step the folder should be a good starting point for @sastels to start adding his code

To-do: An extract_endpoint_integration step will eventually need to be added but I'm not sure what that looks like yet


sastels:

PR #203, opened 2018-02-17
Fail python CI if code coverage is below 80%

buckley-w-david:
This PR adds code coverage check to CI
If code coverage is below 80%, CI will fail.


JasonMWhite:
Looks like this never got a final CI run. :shipit: if 🍏

buckley-w-david:
I believe that it did, it's just displaying the commits in the wrong order in the preview in the Conversations tab. The commit that I added last is the one that has been ran on, and the order is correct in the actual Commits tab.

PR #204, opened 2018-02-18
WIP: Filter on nested documents

sleepycat:
This is the groundwork needed for supporting filters on arbitrary
fields. This also sets us up for painlessly supporting multiple filters.

### DIFF
--- a/api/src/schema/resolvers.js
+++ b/api/src/schema/resolvers.js
@@ -3,9 +3,22 @@ import MongoPaging from 'mongo-cursor-pagination'
 import { GraphQLError } from 'graphql'
 import {
   comparators,
-  createQuery,
   hasMoreThanOneComparator,
 } from '../utilities'
+/* eslint-disable import/named */
+import {
+  dwellingHouseId,
+  dwellingYearBuilt,
+  dwellingCity,
+  dwellingRegion,
+  dwellingForwardSortationArea,
+  ventilationTypeEnglish,
+  ventilationTypeFrench,
+  ventilationAirFlowRateLps,
+  ventilationAirFlowRateCfm,
+  ventilationEfficiency,
+} from './enums'
+/* eslint-enable import/named */
 
 const resolvers = {
   ForwardSortationArea,
@@ -31,7 +44,43 @@ const resolvers = {
         )
       }
 
-      let query = createQuery(forwardSortationArea, filter)
+      let query = {
+        $and: [
+          {
+            forwardSortationArea,
+          },
+        ],
+      }
+
+      // { field: 'yearBuilt', gt: '1990' }
+      if (filter) {
+        for (let key in filter) {
+          // make sure the key exists in the comparators' keys
+          if (Object.keys(comparators).includes(key)) {
+            // if the value is a number
+            // - optionally starting with a minus sign
+            // - containing zero or one decimal places
+            // convert it to a float instead of a string
+            if (filter[key].match(/^-?\d+\.?\d+$/)) {
+              filter[key] = parseFloat(filter[key])
+            }
+
+            // We are evaling our own code here, not user input.
+            // Filter.field is our code, stringified and stored in an enum.
+            // The user choses one of the enum values and we convert the string
+            // back to a function which accepts a matcher and generates the
+            // query we need to find that field.
+            // eslint-disable-next-line no-eval
+            let queryGenerator =  eval(filter.field)
+            let attrQuery = queryGenerator({
+              [comparators[key]]: filter[key],
+            })
+            query['$and'].push(attrQuery)
+          }
+        }
+      }
+
+      // let query = createQuery(forwardSortationArea, filter)
 
       let result = await MongoPaging.find(client, {
         query,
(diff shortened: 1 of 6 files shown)


api/test/enum.test.js
```diff
@@ -0,0 +1,284 @@
...
+  })
+
+  describe('ventilationTypeFrench', () => {
+    it('returns a query object', () => {
+      expect(ventilationTypeFrench('fou')).toEqual({
+        evaluations: {
+          $elemMatch: {
+            ventilations: {
+              $elemMatch: {
+                typeFrench: 'fou',
```
pcraig3: lol

api/test/queries.test.js
```diff
@@ -489,19 +489,130 @@ describe('queries', () => {
...
+                 }
+               }`,
+            })
+          let { dwellings: { results: [first] } } = response.body.data
+          expect(first.city).toEqual('Charlottetown')
+        })
+      })
+
+      describe('lt: less than', () => {
+        it('filters out results where the selected field has a value less than the selected value', async () => {
```
pcraig3: Pretty nitpicky, but I would interpret 'filtering out' a result as removing it from the results.

Ie, if `results = [1,2,3]` and we only want results less than 2, I would say that [2,3] have been filtered out. (ie, `results = [1]`)

Maybe you could say "it keeps/selects results where the selected field...".

api/test/queries.test.js
```diff
@@ -489,19 +489,130 @@ describe('queries', () => {
...
+                  forwardSortationArea: "C1A"
+                  filter: {field: dwellingYearBuilt eq: "1900"}
                  ) {
                    results {
                      yearBuilt
                    }
                  }
                }`,
             })
+
```
pcraig3: Might be good to have a test in here for not finding something as well. 

Like, if we search `filter: {field: dwellingYearBuilt eq: "1899"}`, then we get no results.

api/test/enum.test.js
```diff
@@ -0,0 +1,284 @@
...
+        },
+      })
+    })
+
+    it('has a toString function that produces an eval friendly expression', () => {
+      let str = ventilationEfficiency.toString()
+      let characters = [...str]
+      expect(characters[0]).toEqual('(')
+      expect(characters[characters.length - 1]).toEqual(')')
+    })
```
pcraig3: I'm not sure what's the craic with Jest testing, but in pytest has this notion of parameterized tests, where you can run the same test with lots of different parameters. ([here's an example of testing for dates](https://github.com/alphagov/notifications-api/blob/fbe505a9916aa0f68f70193f51be3654665997cd/tests/app/dao/test_date_utils.py#L20-L31).) 

Is this possible with Jest? These `... eval friendly expression` tests look pretty boilerplate.

api/src/schema/resolvers.js
```diff
@@ -31,7 +41,40 @@ const resolvers = {
...
+              filter[key] = parseFloat(filter[key])
+            }
+
+            // We are evaling our own code here, not user input.
+            // Filter.field is our code, stringified and stored in an enum.
+            // The user choses one of the enum values and we convert the string
+            // back to a function which accepts a matcher and generates the
+            // query we need to find that field.
+            // eslint-disable-next-line no-eval
+            let queryGenerator = eval(filter.field)
```
pcraig3: Ignore me if this is a naive comment, but is there a way to import './enums' as a module with and then refer to methods in it using a string?

Ie, `enums['dwellingHouseId']`, or something?

Because at the minute, _we_ know that `filter.field` is one of the enums imported at the top of the file, but it's not immediately obvious to the casual reader.

api/src/schema/resolvers.js
```diff
@@ -31,7 +41,40 @@ const resolvers = {
...
+            let attrQuery = queryGenerator({
+              [comparators[key]]: filter[key],
+            })
+            query['$and'].push(attrQuery)
```
pcraig3: Looks really clean though, eh? I know we normally frown on `eval`s, but I think this is pretty neat.

sleepycat: You can see how all of a sudden multiple queries are going to be easy. :smile:

api/src/schema/resolvers.js
sleepycat: `import * as enums from './enums'` would give exactly that: `enums['dwellingHouseId']`.
The enums are the only values possible in there according to the type system. I think what you are getting at here is that it could/should be a little more obvious what's going on. I have some ideas for that I am about to start on.

api/test/enum.test.js
sleepycat: Totally possible. You're right that it feels like these could be generated. Now that the pattern is clearer, I'll explore when I refactor.

api/test/queries.test.js
sleepycat: Agree. I'll add that shortly.

api/src/schema/resolvers.js
pcraig3: Yeah, that's what I meant. 👍 if you're working on it.

api/test/enum.test.js
pcraig3: Sounds good.

PR #207, opened 2018-02-19
Log exceptions

buckley-w-david:
This PR adds code to catch exceptions in `transform.run` and log them instead of halting execution.


buckley-w-david:

PR #209, opened 2018-02-19
Azure circle CI setup 

evadb:
Azure circle CI setup to use the emulator for unit tests. The values are provided as dev values by Microsoft to connect to the emulator (no live production keys were harmed in this pull request)


evadb:
A good ol CLI set up PR with some fun commits. @sastels this should set it up for you to be able to access the Azure emulator while running unit tests via circle CI

PR #211, opened 2018-02-19
Extract endpoint

sastels:
Create a microservice for the extractor to push data through. The microservice stores the data in Azure, where the transform-load code will read from.


extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -0,0 +1,68 @@
...
+    if App.config['SECRET_KEY'] == DEFAULT_ENDPOINT_SECRET_KEY:
+        raise ValueError("Need to define environment variable ENDPOINT_SECRET_KEY")
+    if 'signature' not in flask.request.form:
+        flask.abort(404)
+    if 'salt' not in flask.request.form:
+        flask.abort(404)
+    if flask.request.form.get('filename', '') == '':
+        flask.abort(404)
+    if 'file' not in flask.request.files:
+        flask.abort(404)
```
buckley-w-david: I feel like these should maybe be `400` errors, not `404`. It's not that there wasn't anything here, it's that the request is malformed/missing data

buckley-w-david: Also, is there not some flask enum or something for aborting by a name? 
Even though these are pretty well known, it feels a little "magic number"-y. Maybe a variable `BAD_REQUEST` and `NOT_FOUND` that are `400` and `404` respectively?

extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -0,0 +1,68 @@
...
+    file = flask.request.files['file']
+    file_as_string = base64.b64encode(file.read()).decode('utf-8')
+    signature = crypt_utils.sign_string(salt=flask.request.form['salt'], key=App.config['SECRET_KEY'],
+                                        data=file_as_string)
+    if flask.request.form['signature'] != signature:
+        flask.abort(404)
+
+    file.seek(0)
+    filename = utils.secure_filename(flask.request.form['filename'])
+    if azure_utils.upload_stream_to_azure(App.config['AZURE_COORDINATES'], file, filename):
```
buckley-w-david: You're already reading the entire file into memory on `54`, instead of rewinding the stream and re-reading it in upload, do you think it'd be better to just keep the store the read string, and upload that?

extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -0,0 +1,68 @@
...
+        return 'success'
+    return 'failure'
```
buckley-w-david: Seems odd to return strings indicating success or failure when we have perfectly good HTTP status codes for that

extract_endpoint/tests/test_endpoint.py
```diff
@@ -0,0 +1,165 @@
...
+import base64
+import pytest
+from flask import testing
+from azure.storage import blob
+from extract_endpoint import azure_utils
+from extract_endpoint import crypt_utils
+from extract_endpoint import endpoint
+
+
+endpoint.App.testing = True
```
buckley-w-david: Is this still needed? No idea what it's for, I just notice that it doesn't seem to be used anywhere.

sastels: yeah, it configures flask to allow exceptions to flow through to the test_client.  ¯\_(ツ)_/¯

extract_endpoint/tests/test_endpoint.py
```diff
@@ -0,0 +1,165 @@
...
+
+@pytest.fixture
+def sample_salt() -> str:
+    return 'sample salt'
+
+
+@pytest.fixture
+def sample_secret_key() -> str:
+    endpoint.App.config['SECRET_KEY'] = 'sample secret key'
+    return endpoint.App.config['SECRET_KEY']
```
buckley-w-david: This fixture has side effects that it doesn't clean up, I'd save the current set `endpoint.App.config['SECRET_KEY']`, and reset it afterwards

extract_endpoint/tests/test_endpoint.py
```diff
@@ -0,0 +1,165 @@
...
+from extract_endpoint import crypt_utils
+from extract_endpoint import endpoint
+
+
+endpoint.App.testing = True
+
+
+@pytest.fixture
+def azure_service() -> blob.BlockBlobService:
+    return blob.BlockBlobService(account_name=azure_utils.AZURE_EMULATOR_COORDS.account,
```
buckley-w-david: It didn't occur to me until now, but does it make more sense to have the `AZURE_EMULATOR_COORDS` defined in one of the test files, instead of in a src file? Unless I am mistaken, it's only ever used in tests

extract_endpoint/tests/test_endpoint.py
JasonMWhite: yes please. Fixtures/tests should not have any side effects beyond the fixture/test scope. `yield` is your friend here. If it were an environment variable, `monkeypatch.setenv` would do it.

extract_endpoint/src/extract_endpoint/endpoint.py
sastels: 👍 to better status codes. (there doesn't seem to be any constants for them defined in flask)

buckley-w-david: If there are no flask variables for them, I think we should make them ourselves, just some variables that describe the status code

extract_endpoint/tests/test_endpoint.py
sastels: Perhaps, though they're needed in all of the test files in this project, so we'd be importing one test file into the other test files. Seems fugly.

buckley-w-david: It doesn't have to be stored in one of the test files that actually contains tests, it can be in just another .py file in the tests directory that the other tests can import

buckley-w-david: Also a good candidate for a test fixture in `conftest.py`

extract_endpoint/src/extract_endpoint/endpoint.py
sastels: ah ha, `http.HTTPStatus` seems to be useful here.

extract_endpoint/src/extract_endpoint/endpoint.py
sastels: 👍

extract_endpoint/src/extract_endpoint/endpoint.py
sastels: 👍

extract_endpoint/tests/test_endpoint.py
sastels: 👍

extract_endpoint/src/extract_endpoint/endpoint.py
buckley-w-david: Yes it does 👍

extract_endpoint/tests/test_endpoint.py
sastels: 👍

PR #216, opened 2018-02-19
include optional exception message for greater readability

JasonMWhite:
Exactly as the title suggests.

### DIFF
--- a/etl/src/energuide/exceptions.py
+++ b/etl/src/energuide/exceptions.py
@@ -15,6 +15,6 @@ class InvalidInputDataError(EnerguideError):
 
 class InvalidEmbeddedDataTypeError(EnerguideError):
 
-    def __init__(self, data_class: type, *args: typing.Tuple, **kwargs: typing.Dict) -> None:
-        super().__init__(*args, **kwargs)
+    def __init__(self, data_class: type, msg: typing.Optional[str] = None) -> None:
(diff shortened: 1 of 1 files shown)


etl/src/energuide/exceptions.py
```diff
@@ -15,6 +15,6 @@ class InvalidInputDataError(EnerguideError):
 
 class InvalidEmbeddedDataTypeError(EnerguideError):
 
-    def __init__(self, data_class: type, *args: typing.Tuple, **kwargs: typing.Dict) -> None:
-        super().__init__(*args, **kwargs)
+    def __init__(self, data_class: type, msg: typing.Optional[str] = None) -> None:
         self.data_class = data_class
+        super().__init__(msg)
```
buckley-w-david: Should we be throwing away arguments that could have been passed in in `*args` and `**kwargs`?

JasonMWhite: That's a good question. We're already violating the Liskov Substitution Principle, and I'm not sure of the best pattern to follow to fix that.

I can pass in `*args` and `**kwargs` anyway easily enough.

JasonMWhite: Actually, I can't because `msg` can't come after `*args` (since it has a default value of `None`). I'd like to proceed as-is on this one until we have a better idea.

PR #217, opened 2018-02-19
Check that things are returned

sleepycat:
Previously these tests baked the houseId into the expectations. This
wasn't a core part of the tests (given that they are just testing that
the query object is usable by mongo), and assumes an order that appears
not to hold... sometimes...


pcraig3:
Yeah, we have no default ordering so we shouldn't expect any particular id.

PR #219, opened 2018-02-19
add heating element

JasonMWhite:
This PR adds the long-awaited Heating element. It isn't yet included in the actual transform operation, merely the ability to parse it from xml and serialize it with `to_dict`.

There are some known holes that I was not able to yet figure out - no known examples of electric baseboards, P9-11 efficiency heaters, or combo heating/water heater systems. The parsing logic should fail loudly with explicit error messages in these cases.

### DIFF
--- a/etl/src/energuide/embedded/heating.py
+++ b/etl/src/energuide/embedded/heating.py
@@ -1,0 +1,151 @@
+import enum
+import typing
+from energuide import bilingual
+from energuide import element
+from energuide.exceptions import InvalidEmbeddedDataTypeError
+
+
+class HeatingType(enum.Enum):
+    BASEBOARD = enum.auto()
+    FURNACE = enum.auto()
+    BOILER = enum.auto()
+    COMBO_HEAT_DHW = enum.auto()
+    P911 = enum.auto()
+
+
+class EnergySource(enum.Enum):
+    ELECTRIC = enum.auto()
+    NATURAL_GAS = enum.auto()
+    OIL = enum.auto()
+    PROPANE = enum.auto()
+    WOOD = enum.auto()
+
+
+class _Heating(typing.NamedTuple):
+    heating_type: HeatingType
+    energy_source: EnergySource
+    equipment_type: bilingual.Bilingual
+    label: str
+    output_size: float
+    efficiency: float
+    steady_state: str
+
+
+class Heating(_Heating):
+
+    _KWH_TO_BTU = 3412.142
+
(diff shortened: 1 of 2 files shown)


etl/src/energuide/embedded/heating.py
```diff
@@ -0,0 +1,151 @@
...
+        return capacity
+
+    @classmethod
+    def _get_heating_type(cls, node: element.Element) -> HeatingType:
+        candidates = [candidate.tag for candidate in node.xpath('Type1/*')]
+        heating_type: typing.Optional[HeatingType] = None
+        for candidate in candidates:
+            if candidate in cls._HEATING_TYPE_NODE_NAMES:
+                heating_type = cls._HEATING_TYPE_NODE_NAMES[candidate]
+                break
```
buckley-w-david: This section seems odd to me, I'm guessing there's a good reason for it, possibly motivated by lack of test data?

etl/tests/embedded/test_heating.py
```diff
@@ -0,0 +1,174 @@
...
+            </Mode>
+            <Power isCalculated="true" low="0" high="145.5" />
+        </FansAndPump>
+    </Type1>
+    <Type2 shadingInF280Cooling="AccountedFor" />
+</HeatingCooling>
+    """
+    node = element.Element.from_string(data)
+    type1_node: element.Element = node.xpath('Type1')[0]
+    heating_node = sample_heating_node(heating_type)
```
buckley-w-david: This seems odd to have down here when the other node creation calls are done at the top of the function

etl/src/energuide/embedded/heating.py
JasonMWhite: Yes, we don't know what all the heating type node names are. There are multiple possible child nodes of `Type1`, only one of them is the one we want, and we don't know all the names. The three we know are `Baseboards`, `Furnace`, and `Boiler`. I haven't been able to figure out the other 2 yet, so I intentionally want it to fail and raise an error.

etl/tests/embedded/test_heating.py
JasonMWhite: At the top, if they are `None` they get set to their default value. Should I move this line up there with them for consistency, although it's slightly different?

buckley-w-david: I'll leave it up to you, if I were writing it I'd put it at the top with the others, but that seems more like an personal preference thing

PR #220, opened 2018-02-19
Refactor filters: {field, comparator, value}

sleepycat:
This significantly simplifies the code.


api/src/schema/resolvers.js
```diff
@@ -50,30 +42,21 @@ const resolvers = {
...
-            // eslint-disable-next-line no-eval
-            let queryGenerator = eval(filter.field)
-            let attrQuery = queryGenerator({
-              [comparators[key]]: filter[key],
-            })
-            query['$and'].push(attrQuery)
-          }
+        let value = filter.value
+        if (filter.value.match(/^-?\d+\.?\d+$/)) {
+          value = parseFloat(filter.value)
```
pcraig3: I think it's worth keeping in the comment describing the regex, in one form or another.

api/src/utilities.js
```diff
@@ -1,22 +1,3 @@
-// Mongodb filter map:
-export const comparators = {
-  gt: '$gt',
-  lt: '$lt',
-  eq: '$eq',
-}
-
-export function hasMoreThanOneComparator(filter) {
```
pcraig3: Yeah, getting rid of this is a win 🥇

api/test/queries.test.js
```diff
@@ -503,15 +503,15 @@ describe('queries', () => {
       })
 
       describe('lt: less than', () => {
-        it('filters out results where the selected field has a value less than the selected value', async () => {
+        it('returns dwellings where the field is less than the given value', async () => {
```
pcraig3: There are a few tests that use the `'filters out results'` description string, so best rename them all.

PR #223, opened 2018-02-20
[Windows] Update README

buckley-w-david:
This PR introduces a small note to the windows README asking users to make sure they are adding the libraries to the correct 'PATH' variable.


README.md
```diff
@@ -107,6 +107,8 @@ node --version
 ```
 If version is not shown, the path to the library has to be added to the 'PATH' environment variable in System Properties.
 
+NOTE: If you're having problems, make sure that you are not adding the library to a different users 'PATH' (such as the user path of the local admin). The library can be added to the System 'PATH' variable so that it is visible to all users on the machine.
```
JasonMWhite: Can we include a screenshot?

buckley-w-david: I was thinking that, but was unsure if we wanted to go that far, I'll get one added in.

pcraig3: Yeah, that would be good -- even if it were just a link.

pcraig3: Also, small typo.

>  a different users 'PATH'

should be 

>  a different user's 'PATH'

PR #224, opened 2018-02-20
Windows compatibility changes

buckley-w-david:
This PR introduces small refactoring/changes to the ETL code to make it compatible with development on a Windows machine.

1. If an `OverflowError` occurs when setting the max field size, fall back to the size of a `long` in C in windows
2. Open the csv file with UTF-8 encoding explicitly (instead of relying on locale default)
3. Delete invisible characters at the beginning of `sample.h2k`

ping @pcraig3 and @vadjr to let you know I believe this resolves all the currently known incompatibilities with Windows and Unix.

### DIFF
--- a/etl/tests/sample.h2k
+++ b/etl/tests/sample.h2k
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="UTF-8" ?>
+<?xml version="1.0" encoding="UTF-8" ?>
 <HouseFile xml:lang="en" uiUnits="Imperial">
     <Version major="1" minor="3">
         <Labels>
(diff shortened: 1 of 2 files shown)


etl/tests/sample.h2k
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="UTF-8" ?>
+<?xml version="1.0" encoding="UTF-8" ?>
```
buckley-w-david: There seemed to be 3 invisible characters at the start of this line that `lxml` was able to ignore on unix based systems, but caused a crash on windows.

pcraig3: O_o

Good job tracking that one down.

PR #225, opened 2018-02-20
Add floor fields to filter field enum

sleepycat:
This allows users to filter results based on values on floor type.
Additionally, the tests have been rethought and refactored so they can just be generated.

### DIFF
--- a/api/src/schema/enums.js
+++ b/api/src/schema/enums.js
@@ -35,7 +35,9 @@ const ventilationFields = [
 
 ventilationFields.forEach(attr => {
   // eslint-disable-next-line no-new-func
-  let fn = new Function('matcher', `
+  let fn = new Function(
+    'matcher',
+    `
   return {
     evaluations: {
       $elemMatch: {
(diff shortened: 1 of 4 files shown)


api/src/schema/enums.js
```diff
@@ -45,6 +47,39 @@ ventilationFields.forEach(attr => {
       },
     },
   }
-  `)
+  `,
+  )
   module.exports[generateName('ventilation', attr)] = attachToString(fn)
 })
+
+// The fields on the ventilation type
```
pcraig3: Should read `// The fields on the floor type`

api/test/enum.test.js
```diff
@@ -34,251 +23,74 @@ describe('Enum values', () => {
...
+    },
+  }
+
+  Object.keys(testData).forEach(functionName => {
+    describe(functionName, () => {
+      it('returns a query object capable of returning data', async () => {
+        let { testValue } = testData[functionName] // eslint-disable-line
+        let query = enumFunctions[functionName](testValue) //eslint-disable-line
+        let result = await collection.findOne(query)
+        expect(result).not.toBe(null)
```
pcraig3: Much cleaner than before 👍 👍

sleepycat: Emergent order!

PR #226, opened 2018-02-20
Fix regex to handle integers and floats

sleepycat:
This regex is responsible for converting strings containing numbers into
actual numbers so that values passed to mongo will have the correct
type. As the number of types you can search by expand, it turns out that
this regex doesn't handle integers. This changes that.

### DIFF
--- a/api/src/schema/resolvers.js
+++ b/api/src/schema/resolvers.js
@@ -57,7 +57,7 @@ const resolvers = {
           // - optionally starting with a minus sign
           // - containing zero or one decimal places
           // convert it to a float instead of a string
-          if (filter.value.match(/^-?\d+\.?\d+$/)) {
+          if (filter.value.match(/^-?\d+$|^-?\d+\.\d+$/)) {
             value = parseFloat(filter.value)
           }
           // We are evaling our own code here, not user input.
--- a/api/test/queries.test.js
+++ b/api/test/queries.test.js
@@ -555,11 +555,11 @@ describe('queries', () => {
               query: `{
                 dwellingsInFSA(
                   forwardSortationArea: "C1A"
-                  filter: {
+                  filters: [{
                     field: dwellingYearBuilt
                     comparator: gt
                     value: "1"
-                  }) {
+                  }]) {
                   results {
                     yearBuilt
                   }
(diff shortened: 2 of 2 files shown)


api/src/schema/resolvers.js
```diff
@@ -43,11 +43,9 @@ const resolvers = {
 
       if (filter) {
         let value = filter.value
-        // if the value is a number
-        // - optionally starting with a minus sign
-        // - containing zero or one decimal places
-        // convert it to a float instead of a string
-        if (filter.value.match(/^-?\d+\.?\d+$/)) {
```
pcraig3: Oh yeah, this won't match a single digit. 😖Good catch.

api/src/schema/resolvers.js
```diff
@@ -43,11 +43,9 @@ const resolvers = {
...
+        // if the value is either an integer or a float, optionally starting
+        // with a minus sign convert it to a float instead of a string
+        if (filter.value.match(/^-?[0-9]+$|^-?\d+\.?[0-9]*$/)) {
```
pcraig3: Two things:

1. we should probably standardize on either `[0-9]` or `\d` for digits and make them all this way.

2. the problem with the last half of the expression (`^-?\d+\.?[0-9]*`) is that if you make the first digit optional, you would allow values like `-.1` or `.1`, whereas if you make the trailing digit optional, you would allow values like `-1.` or `1.`. Having said that, `parseFloat` seems to work fine with either. 

***

I don't know if the first part of the OR is doing anything important. If we just changed the original expression to `/^-?\d+\.?\d*$/`, we would have the same result.

My opinion: just change the original expression to either  `/^-?\d*\.?\d+$/` (accepts `-.1`) or  `/^-?\d+\.?\d*$/`  (accepts `-1.`) and write a test that confirms correct behaviour.

api/src/schema/resolvers.js
```diff
@@ -45,7 +45,7 @@ const resolvers = {
         let value = filter.value
         // if the value is either an integer or a float, optionally starting
         // with a minus sign convert it to a float instead of a string
-        if (filter.value.match(/^-?[0-9]+$|^-?\d+\.?[0-9]*$/)) {
+        if (filter.value.match(/^-?\d+$|^-?\d+\.\d+$/)) {
```
pcraig3: Right, yes, and then `1` or `-1` is caught by the first half. 👍

PR #228, opened 2018-02-20
load heating data

JasonMWhite:
This PR takes the newly-added `Heating` element and adds it to the data to be loaded to MongoDB.

Fixed a couple of broken windows on the way.

### DIFF
--- a/etl/src/energuide/dwelling.py
+++ b/etl/src/energuide/dwelling.py
@@ -7,6 +7,7 @@ from energuide import validator
 from energuide.embedded import ceiling
 from energuide.embedded import code
 from energuide.embedded import floor
+from energuide.embedded import heating
 from energuide.embedded import wall
 from energuide.embedded import door
 from energuide.embedded import window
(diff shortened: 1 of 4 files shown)


etl/src/energuide/dwelling.py
```diff
@@ -90,12 +91,12 @@ class _ParsedDwellingDataRow(typing.NamedTuple):
...
     water_heatings: typing.List[water_heating.WaterHeating]
     ventilations: typing.List[ventilation.Ventilation]
-
+    heating_system: heating.Heating
 
 
 class ParsedDwellingDataRow(_ParsedDwellingDataRow):
 
-    _XML_SCHEMA = {'type': 'xml', 'coerce': 'parse_xml'}
+    _XML_SCHEMA = {'type': 'xml', 'required': True, 'coerce': 'parse_xml'}
```
JasonMWhite: Don't know why this wasn't always the case.

etl/src/energuide/embedded/heating.py
```diff
@@ -82,11 +82,11 @@ def _get_output_size(cls, node: element.Element) -> float:
         capacity: typing.Optional[float]
         if units == 'kW':
             capacity = capacity_value
-        elif units == 'btu/h':
+        elif units == 'btu/hr':
             capacity = capacity_value / cls._KWH_TO_BTU
         else:
             raise InvalidEmbeddedDataTypeError(
-                Heating, f'Unknown capacity units: f{units}')
+                Heating, f'Unknown capacity units: {units}')
```
JasonMWhite: The above two bugs had me googling for what the heck a `fbtu/hr` was. It showed up in my stacktrace, so yay logging!

etl/tests/test_dwelling.py
```diff
@@ -564,8 +565,8 @@ def test_from_row(self, sample_input_d: reader.InputData) -> None:
             ventilations=[
                 ventilation.Ventilation(
                     ventilation_type=ventilation.VentilationType.NOT_ENERGY_STAR_NOT_INSTITUTE_CERTIFIED,
-                    air_flow_rate=220,
-                    efficiency=55,
+                    air_flow_rate=220.0,
+                    efficiency=55.0,
```
JasonMWhite: Technically `==`, but shows up as a difference in the output string when other things are different.

PR #234, opened 2018-02-22
Add field descriptions/translations for most of our API fields

pcraig3:
This PR adds documentation to most of our API fields. 

For example, the following query will return the description of what a ventilation system is:
```
{
  __type(name: "Ventilation") {
    name
    description
  }
}
```

Currently this does not include translations for 
- [our filter fields](https://github.com/cds-snc/nrcan_api/blob/b577d555cf46b5570ce661e6d81b956b73d37e3d/api/src/schema/index.js#L164-L184) (generated fields used to filter our results) 
- our Scalar values (like the [ForwardSortationArea](https://github.com/cds-snc/nrcan_api/blob/master/api/src/schema/types/ForwardSortationArea.js#L11))

Making that into a future pull request as this one is already pretty large.

***

**Note on the documentation**

- The majority of field descriptions are identical or very similar to something that exists in the NRCAN documentation I looked through
- I have tried to be descriptive where possible. (ie, rather than saying `glazingType` is `details about glazing`, I’ve tried to find a description of what glazing means: `Number of panes of transparent material in a window`
- I’ve tried to include the full descriptions of units (`Litres per second` instead of just `lps`)
- I haven’t included options for various fields (ie, windowType can be “Wood”, “Aluminum”, etc etc)
- I’ve used `(en)` and `(fr)` in the descriptions when fields have different languages
- I’ve removed most cases of "the" and passive sentence constructions

I don’t know if we have a style guide or anything approaching, but my overriding principle was clarity while sticking close to the NRCAN source documentation.

French translations: I’ve put in what I consider to be placeholder translations. As in, they are better than nothing but it would be ideal if someone looked them over.

Ways to contribute: 
- If you would like to standardize the English documentation based on some principle, I am all for it
- If you would like to rewrite the French translations, that would be very welcome


pcraig3:
Talked to @annieleblond in real life and she says she can have a look in the next few days. Merging this now and then we'll iterate.

annieleblond:
As Translation if they could to it, since there is quite a lot. At a glance, the FR seemed pretty good, so all they will have to do is confirmed. Waiting for their answer.

PR #240, opened 2018-02-22
Always generate floors

buckley-w-david:
This PR addresses an issue that I noticed regarding how foundation parsing should deal with missing elements.

Report Generator always generates floor rows, even if floor tag is missing, whereas the lack of walls and headers will result in no rows. This PR now emulates that behavior

### DIFF
--- a/etl/src/energuide/embedded/basement.py
+++ b/etl/src/energuide/embedded/basement.py
@@ -166,13 +166,13 @@ class BasementFloor(_BasementFloor):
 
     @classmethod
     def from_crawlspace(cls, floor: typing.Optional[element.Element]) -> typing.List['BasementFloor']:
-        return [
-            cls._from_data(floor, 'AddedToSlab', FloorType.SLAB)
-            if floor is not None else cls._empty_floor(FloorType.SLAB),
-
-            cls._from_data(floor, 'FloorsAbove', FloorType.FLOOR_ABOVE_CRAWLSPACE)
-            if floor is not None else cls._empty_floor(FloorType.FLOOR_ABOVE_CRAWLSPACE),
-        ]
+        if floor is None:
+            return [cls._empty_floor(FloorType.SLAB), cls._empty_floor(FloorType.FLOOR_ABOVE_CRAWLSPACE)]
+        else:
+            return [
+                cls._from_data(floor, 'AddedToSlab', FloorType.SLAB),
+                cls._from_data(floor, 'FloorsAbove', FloorType.FLOOR_ABOVE_CRAWLSPACE),
+            ]
 
     @classmethod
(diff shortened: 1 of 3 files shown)


buckley-w-david:
[Extract/Load Foundation data](https://trello.com/c/MHCxw57a/107-extract-load-foundation-data)

etl/src/energuide/embedded/basement.py
```diff
@@ -104,7 +104,23 @@ class BasementFloor(_BasementFloor):
     }
 
     @classmethod
-    def _from_data(cls, floor: element.Element, construction_type: str, floor_type: FloorType) -> 'BasementFloor':
+    def _from_data(cls,
+                   floor: typing.Optional[element.Element],
+                   construction_type: str,
+                   floor_type: FloorType) -> 'BasementFloor':
+
+        if floor is None:
```
JasonMWhite: This is weird. `construction_type` isn't used if `floor` is None - hints that maybe this should be a different function.

buckley-w-david: Do you think some kind of `_empty_floor` function would be a better idea, that is called instead of this when `floor` in `None`?

etl/src/energuide/embedded/basement.py
```diff
@@ -141,19 +158,28 @@ def _from_data(cls, floor: element.Element, construction_type: str, floor_type:
...
-    def from_crawlspace(cls, floor: element.Element) -> typing.List['BasementFloor']:
+    def from_crawlspace(cls, floor: typing.Optional[element.Element]) -> typing.List['BasementFloor']:
         return [
-            cls._from_data(floor, 'AddedToSlab', FloorType.SLAB),
-            cls._from_data(floor, 'FloorsAbove', FloorType.FLOOR_ABOVE_CRAWLSPACE),
+            cls._from_data(floor, 'AddedToSlab', FloorType.SLAB)
+            if floor is not None else cls._empty_floor(FloorType.SLAB),
+
+            cls._from_data(floor, 'FloorsAbove', FloorType.FLOOR_ABOVE_CRAWLSPACE)
+            if floor is not None else cls._empty_floor(FloorType.FLOOR_ABOVE_CRAWLSPACE),
```
JasonMWhite: I think this would be easier to read if you reversed the logic:
```
if floor is None:
    return [cls._empty_floor(FloorType.SLAB), cls._empty_floor(FloorType.FLOOR_ABOVE_CRAWLSPACE)]
else:
    return [
        cls._from_data(floor, 'AddedToSlab', FloorType.SLAB),
        cls._from_data(floor, 'FloorsAbove', FloorType.FLOOR_ABOVE_CRAWLSPACE),
    ]
```

PR #241, opened 2018-02-22
Drop collection before load

buckley-w-david:
This PR adds functionality to drop the collection by default before loading

### DIFF
--- a/etl/src/energuide/cli.py
+++ b/etl/src/energuide/cli.py
@@ -23,7 +23,15 @@ def main() -> None:
               envvar=database.EnvVariables.collection.value,
               default=database.EnvDefaults.collection.value)
 @click.option('--filename', type=click.Path(exists=True), required=True)
-def load(username: str, password: str, host: str, port: int, db_name: str, collection: str, filename: str) -> None:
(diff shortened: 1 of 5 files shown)


etl/src/energuide/cli.py
```diff
@@ -23,7 +23,15 @@ def main() -> None:
               envvar=database.EnvVariables.collection.value,
               default=database.EnvDefaults.collection.value)
 @click.option('--filename', type=click.Path(exists=True), required=True)
-def load(username: str, password: str, host: str, port: int, db_name: str, collection: str, filename: str) -> None:
+@click.option('-a', '--append', is_flag=True, help='Sets load to append instead overwrite')
```
JasonMWhite: Defaults to `False`?

JasonMWhite: Can we update the `help` statement to something like `Append data instead of overwriting`

buckley-w-david: It's a flag, so it'd be called like `energuide load --append --filename filename.zip`, not `energuide load --append True --filename filename.zip`. It's about if it is present or not, so lack of argument means `False`.

PR #246, opened 2018-02-23
Slack notification

buckley-w-david:
This PR adds a notification to the NRCan slack channel when the master branch fails CI.

Currently only added to the `etl` job while the exact message and mechanism is worked out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,9 +82,6 @@ jobs:
               exit 1
             fi
       - run:
-          name: Fail
-          command: exit 1
-      - run:
           name: Report failing Master
           command: ~/repo/scripts/notify_failure.sh
           when: on_fail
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -81,6 +81,13 @@ jobs:
               echo Failing build due to less than 80% test coverage
               exit 1
             fi
+      - run:
+          name: Report failing Master
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "master" ]; then
```
JasonMWhite: we should make this an executable script in `./scripts/notify_failure.sh` to simplify adding it to more job nodes.

JasonMWhite:
It's unfortunate we have to do it at this level, but I don't currently see a better option.

scripts/notify_failure.sh
```diff
@@ -0,0 +1,3 @@
+if [ "${CIRCLE_BRANCH}" == "slack_notification" ]; then
```
JasonMWhite: Make sure you `chmod` this file to make it executable.

PR #249, opened 2018-02-23
Foundation exceptions

buckley-w-david:
This PR adds exception handling to the basements type, with accompanying tests.

Length is larger than I'd like, but the majority of it is spent on test fixture data.


etl/src/energuide/embedded/basement.py
```diff
@@ -64,12 +64,17 @@ class BasementHeader(_BasementHeader):
...
+            return BasementHeader(
+                nominal_insulation=insulation.Insulation(
+                    float(header.xpath('Construction/Type/@nominalInsulation')[0])
+                ),
+                effective_insulation=insulation.Insulation(float(header.xpath('Construction/Type/@rValue')[0])),
+                height=distance.Distance(float(header.xpath('Measurements/@height')[0])),
+                perimeter=distance.Distance(float(header.xpath('Measurements/@perimeter')[0])),
+            )
+        except (ValueError, IndexError) as exc:
+            raise InvalidEmbeddedDataTypeError(BasementHeader) from exc
```
JasonMWhite: Please add a meaningful exception message here to make it easier to Cmd-f to an answer.

etl/src/energuide/embedded/basement.py
```diff
@@ -225,49 +236,59 @@ class BasementWall(_BasementWall):
 
     @classmethod
     def _from_data(cls, wall: element.Element, wall_perimiter: float, wall_height: float, tag: WallType):
```
JasonMWhite: Missing a return type

etl/src/energuide/embedded/basement.py
```diff
@@ -349,14 +370,25 @@ def from_data(cls, basement: element.Element) -> 'Basement':
...
-            label=basement.get_text('Label'),
-            configuration_type=basement.xpath('Configuration/@type')[0],
-            walls=walls,
-            floors=floors,
-            header=header,
-        )
+        try:
+            configuration_type = basement.get('Configuration/@type', str)
+            assert len(configuration_type) > 1
+        except AssertionError as exc:
```
JasonMWhite: This is a roundabout way of doing this. Easier:
```
if not configuration_type:
    raise InvalidEmbeddedDataTypeError(Basement, 'Missing configuration type')
```

etl/src/energuide/embedded/basement.py
```diff
@@ -349,14 +370,25 @@ def from_data(cls, basement: element.Element) -> 'Basement':
...
+        try:
+            return Basement(
+                foundation_type=foundation_type,
+                label=basement.get_text('Label'),
+                configuration_type=configuration_type,
+                walls=walls,
+                floors=floors,
+                header=header,
+            )
+        except ElementGetValueError as exc:
```
JasonMWhite: Would be better to combine all the possible exception-raising logic into a single try/except block

etl/src/energuide/embedded/basement.py
```diff
@@ -122,40 +128,45 @@ def _from_data(cls,
...
+                perimeter = (2 * length) + (2 * width)
+                floor_area = length * width
+            else:
+                floor_area = floor.get('Measurements/@area', float)
+                perimeter = floor.get('Measurements/@perimeter', float)
+
+            nominal_insulation = floor.xpath(f'Construction/{construction_type}/@nominalInsulation')
+            effective_insulation = floor.xpath(f'Construction/{construction_type}/@rValue')
+
+            return BasementFloor(
```
JasonMWhite: I don't think this needs to be part of the try block

etl/src/energuide/embedded/basement.py
buckley-w-david: It would look nicer, but would result in less granular error messages.

etl/src/energuide/embedded/basement.py
```diff
@@ -225,49 +236,59 @@ class BasementWall(_BasementWall):
...
-            nominal_insulation=insulation.Insulation(float(wall.attrib['nominalRsi'])),
-            effective_insulation=insulation.Insulation(float(wall.attrib['rsi'])),
-            composite_percentage=percentage,
-            wall_area=area.Area(wall_perimiter * wall_height * (percentage / 100))
-        )
+        try:
+            percentage = float(wall.attrib['percentage'])
+            return BasementWall(
+                wall_type=tag,
+                nominal_insulation=insulation.Insulation(float(wall.attrib['nominalRsi'])),
```
JasonMWhite: this could be `wall.get('@nominalRsi', float)`

etl/src/energuide/embedded/basement.py
JasonMWhite: also: `wall_perimeter`

etl/src/energuide/embedded/basement.py
```diff
@@ -225,49 +236,59 @@ class BasementWall(_BasementWall):
...
+                          for wall_section in interior_wall_sections])
 
-        walls.extend([BasementWall._from_data(wall_section, wall_perimiter, wall_height, WallType.EXTERIOR)
-                      for wall_section in exterior_wall_sections])
+            walls.extend([BasementWall._from_data(wall_section, wall_perimiter, wall_height, WallType.EXTERIOR)
+                          for wall_section in exterior_wall_sections])
 
-        walls.extend([BasementWall._from_data(wall_section, wall_perimiter, pony_height, WallType.PONY)
-                      for wall_section in pony_wall_sections])
+            walls.extend([BasementWall._from_data(wall_section, wall_perimiter, pony_height, WallType.PONY)
```
JasonMWhite: These don't need to be part of the try/except

etl/tests/embedded/test_basement.py
```diff
@@ -665,3 +667,252 @@ def test_basement_to_dict(sample_basement_element: element.Element) -> None:
...
+    node = element.Element.from_string(bad_xml)
+    with pytest.raises(InvalidEmbeddedDataTypeError) as excinfo:
+        basement.BasementFloor.from_crawlspace(node)
+
+    assert excinfo.value.data_class is basement.BasementFloor
+
+
+
+@pytest.mark.parametrize("bad_xml", BAD_FOUNDATION_CRAWLSPACE_WALL_XML)
+def test_bad_crawlspace_wall(bad_xml: str) -> None:
```
JasonMWhite: Better way to do this:
- define a function that return a list of tuples of (Bad XML data, data_class)
- define a single test function that uses this function to parametrize the test
- run the first element of the parametrized tuple through the test, assert that the resulting data_class is the second element

etl/src/energuide/embedded/basement.py
JasonMWhite: My point is that it includes a lot of stuff that couldn't raise these errors, which is undesirable if possible to avoid.

etl/src/energuide/embedded/basement.py
```diff
@@ -142,31 +146,36 @@ def _from_data(cls,
...
-                rectangular=rectangular,
-                nominal_insulation=insulation.Insulation(float(nominal_insulation[0]))
-                if nominal_insulation else None,
+            effective_insulation=insulation.Insulation(effective_insulation)
+            if effective_insulation is not None else None,
 
-                effective_insulation=insulation.Insulation(float(effective_insulation[0]))
-                if effective_insulation else None,
+            length=distance.Distance(length)
+            if length is not None else None,
```
JasonMWhite: Please combine to one line

etl/src/energuide/embedded/basement.py
```diff
@@ -142,31 +146,36 @@ def _from_data(cls,
...
 
-                length=distance.Distance(length)
-                if length is not None else None,
+            width=distance.Distance(width)
+            if width is not None else None,
```
JasonMWhite: 1 line this

PR #250, opened 2018-02-23
typed get function for Element

JasonMWhite:
This PR adds a dynamically-typed `get` function to `Element`. `mypy` is able to determine that if you call `result = my_element.get('/path/to/@attribute', int)`, it is supposed to return an `int`.

This should make it possible to simplify some of the exception handling in #249

### DIFF
--- a/etl/src/energuide/element.py
+++ b/etl/src/energuide/element.py
@@ -1,6 +1,6 @@
 import typing
 from lxml import etree
-from energuide.exceptions import EnerguideError
+from energuide.exceptions import EnerguideError, ElementGetValueError
 
 
 class MalformedXmlError(EnerguideError):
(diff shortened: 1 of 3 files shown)


etl/src/energuide/element.py
```diff
@@ -63,3 +66,15 @@ def tag(self) -> str:
 
     def insert(self, index: int, element: 'Element') -> None:
         self.__node.insert(index, element.__node)
+
+    def get(self, xpath: str, type_: typing.Type[T]) -> T:
+        try:
+            value = self.__node.xpath(xpath)[0]
+        except IndexError as ex:
+            raise EnerguideError(f"Couldn't find element at {xpath} in {self.tag}", ex)
```
JasonMWhite: Should we define a custom `EnerguideError` subtype instead? Where should it live?

buckley-w-david: I'm not sure either way, what's your opinion?

JasonMWhite: I think we should, it will add slightly more context.

PR #253, opened 2018-02-26
[WIP] Upgrade settings

buckley-w-david:
This PR is to show newly added upgrades data.
WIP due to the fact that this is actually 2 or 3 PRs worth of work, will be broken up.

PR #255, opened 2018-02-26
Load upgrades

buckley-w-david:
This PR contains the loading components of #253


etl/tests/test_dwelling.py
```diff
@@ -464,6 +465,16 @@ def raw_codes() -> typing.Dict[str, typing.List[str]]:
...
 
+@pytest.fixture
+def upgrades_input() -> typing.List[str]:
+    doc = [
+        '<Ceilings cost="0" priority="12" />',
+        '<MainWalls cost="1" priority="2" />',
+        '<Foundation cost="2" priority="3" />',
+        ]
+
+    return doc
```
sastels: could just return the list directly.

PR #260, opened 2018-02-26
Use Element.get()

sastels:
Change code such as
```
float(ceiling.xpath('Construction/CeilingType/@nominalInsulation')[0])
```
into
```
ceiling.get('Construction/CeilingType/@nominalInsulation', float)
```


etl/src/energuide/embedded/heating.py
```diff
@@ -120,15 +120,15 @@ def _get_equipment_type(cls, node: element.Element) -> bilingual.Bilingual:
...
     @classmethod
     def from_data(cls, node: element.Element) -> 'Heating':
         return Heating(
             label=node.get_text('Label'),
             output_size=float(Heating._get_output_size(node)),
-            efficiency=float(node.xpath('Type1/*/Specifications/@efficiency')[0]),
+            efficiency=node.get('Type1/*/Specifications/@efficiency', float),
             steady_state=cls._get_steady_state(node),
             heating_type=cls._get_heating_type(node),
             energy_source=cls._get_energy_source(node),
```
sastels: `heating.py` didn't have a try / except block that passed the error along as an `InvalidEmbeddedDataTypeError`. Should we add that?

buckley-w-david: I am aware, it was/is on my to-do list

etl/src/energuide/embedded/ceiling.py
```diff
@@ -28,13 +28,13 @@ def from_data(cls, ceiling: element.Element) -> 'Ceiling':
...
                 effective_insulation=insulation.Insulation(
-                    float(ceiling.xpath('Construction/CeilingType/@rValue')[0])),
-                ceiling_area=area.Area(float(ceiling.xpath('Measurements/@area')[0])),
-                ceiling_length=distance.Distance(float(ceiling.xpath('Measurements/@length')[0])),
+                    ceiling.get('Construction/CeilingType/@rValue', float)),
+                ceiling_area=area.Area(ceiling.get('Measurements/@area', float)),
+                ceiling_length=distance.Distance(ceiling.get('Measurements/@length', float)),
             )
-        except (IndexError, ValueError, AssertionError) as exc:
+        except (ElementGetValueError, AssertionError) as exc:
```
JasonMWhite: If you replace `element.get_text` with `element.get('...', str)` you can also do away with the `AssertionError`. Alternatively (perhaps better) we could update the `assert` with an explicit raise of `ElementGetValueError`.

PR #263, opened 2018-02-26
Use cerberus development version

buckley-w-david:
This PR changes the version of cerberus from the current release (1.1), to a fork of the current development version forked into the cds repo so that it remains static.

This is required because the current release of cerberus contains a bug where coercion does not honour the `nullable` option, meaning even if a key is allowed to be `None`, if the coercion throws an exception on it the validation will fail.

This is fixed in the development (and thus the cds fork) version.

### DIFF
--- a/etl/requirements.txt
+++ b/etl/requirements.txt
@@ -7,7 +7,7 @@ mypy==0.560
 click==6.7
 dnspython==1.15.0
 pymongo==3.6.0
-cerberus==1.1
+-e git+git://github.com/cds-snc/cerberus.git#egg=cerberus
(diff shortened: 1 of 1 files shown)


etl/requirements.txt
```diff
@@ -7,7 +7,7 @@ mypy==0.560
 click==6.7
 dnspython==1.15.0
 pymongo==3.6.0
-cerberus==1.1
+-e git+git://github.com/cds-snc/cerberus.git#egg=cerberus
```
JasonMWhite: Should work without `-e`, which would be preferable.

PR #265, opened 2018-02-27
Post to endpoint

sastels:
Will be used by `extract` to post zipped json files to the endpoint.


extract_endpoint/tests/test_post_to_endpoint.py
```diff
@@ -0,0 +1,127 @@
...
+import psutil
+import pytest
+import _pytest
+import requests
+from azure.storage import blob
+from extract_endpoint import post_to_endpoint, azure_utils
+
+
+class NamedStream(io.BytesIO):
+    name: str
```
sastels: This was required to deal with a python bug that was just recently fixed: https://github.com/python/typeshed/issues/1790

buckley-w-david: That is weird and annoying

sastels: :( indeed. At least it'll be fixed in 3.7

PR #266, opened 2018-02-27
Deploy Endpoint, Add Dockerfile

szinck1:
This PR adds a job called deploy_endpoint to CircleCI which will build a docker image and push it to Docker Hub on each commit to master.


extract_endpoint/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3
+MAINTAINER Steven Zinck <steven.zinck@cds-snc.ca>
+LABEL Description="Python ETL" Vendor="Canadian Digital Service"
```
sastels: The description should be something like "Python ETL Endpoint"

PR #271, opened 2018-02-27
Add file_id to Evaluations

buckley-w-david:
This PR adds the add the file_id field to evaluations, this is needed so that home owners can request information on their evaluation

ping @pcraig3


pcraig3:
belated 👍

PR #272, opened 2018-02-27
Add timestamp to post endpoint

evadb:
This PR intends to add timestamp as a click option with the default being datetime.now(), which is then passed to post_stream and ultimately the end point to upload a timestamp.txt file with the passed timestamp.

Opinions please!

### DIFF
--- a/extract_endpoint/src/extract_endpoint/endpoint.py
+++ b/extract_endpoint/src/extract_endpoint/endpoint.py
@@ -63,7 +63,7 @@ def upload_file() -> typing.Tuple[str, int]:
         flask.abort(HTTPStatus.BAD_REQUEST)
     if 'file' not in flask.request.files:
         flask.abort(HTTPStatus.BAD_REQUEST)
-    if flask.request.form.get('timestamp', None) is None:
+    if 'timestamp' not in flask.request.form:
         flask.abort(HTTPStatus.BAD_REQUEST)
 
     file = flask.request.files['file']
@@ -79,7 +79,7 @@ def upload_file() -> typing.Tuple[str, int]:
     if not azure_utils.upload_bytes_to_azure(App.config['AZURE_COORDINATES'], file_contents, filename):
         flask.abort(HTTPStatus.BAD_GATEWAY)
 
-    timestamp = flask.request.form.get('timestamp', None)
+    timestamp = flask.request.form['timestamp']
 
     if not azure_utils.upload_bytes_to_azure(App.config['AZURE_COORDINATES'], timestamp.encode(), TIMESTAMP_FILENAME):
         flask.abort(HTTPStatus.BAD_GATEWAY)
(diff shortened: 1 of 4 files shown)


extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -63,6 +63,8 @@ def upload_file() -> typing.Tuple[str, int]:
         flask.abort(HTTPStatus.BAD_REQUEST)
     if 'file' not in flask.request.files:
         flask.abort(HTTPStatus.BAD_REQUEST)
+    if flask.request.form.get('timestamp', None) == '':
```
sastels: should compare to `None` rather than ''

extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -78,11 +80,9 @@ def upload_file() -> typing.Tuple[str, int]:
         flask.abort(HTTPStatus.BAD_GATEWAY)
 
     timestamp = flask.request.form.get('timestamp', None)
```
sastels: You can just use `flask.request.form['timestamp']` here since you've already checked for it above.

extract_endpoint/tests/test_post_to_endpoint.py
```diff
@@ -90,9 +96,13 @@ def test_post_stream(azure_service: blob.BlockBlobService,
...
+                     sample_filename: str,
+                     sample_timestamp: str) -> None:
 
-    post_return = post_to_endpoint.post_stream(stream=sample_stream, filename=sample_filename, url=upload_url)
+    post_return = post_to_endpoint.post_stream(stream=sample_stream,
+                                               filename=sample_filename,
+                                               url=upload_url,
+                                               timestamp=sample_timestamp)
     assert post_return.status_code == HTTPStatus.CREATED
     check_file_in_azure(azure_service, azure_emulator_coords, sample_filename, sample_stream_content)
```
sastels: you should also use `check_file_in_azure` to make sure the timestamp file is there. (and also below)

extract_endpoint/src/extract_endpoint/endpoint.py
buckley-w-david: Also no need for the `None` in the `get` call, that's the default value anyway.

extract_endpoint/tests/test_post_to_endpoint.py
```diff
@@ -80,6 +80,11 @@ def sample_timestamp() -> str:
     return datetime.datetime(2013, 1, 1, 0, 0, 0).strftime("%Y-%m-%d %H:%M:%S")
 
 
+@pytest.fixture
+def sample_timestamp_filename() -> str:
+    return "timestamp.txt"
```
sastels: You should probably either return `endpoint.TIMESTAMP_FILENAME` here or set `endpoint.TIMESTAMP_FILENAME` here. Otherwise the file used by the endpoint could differ from the one in this test file.

PR #274, opened 2018-02-28
Start modularizing the schema

sleepycat:
This commit is the beginning of the process of splitting this schema up
into smaller pieces. The Header type (needed for foundations) is
included here as it's the type that touched off this little refactoring
operation, but it's not used or visible anywhere.

### DIFF
--- a/api/src/schema/Header.js
+++ b/api/src/schema/Header.js
@@ -1,0 +1,27 @@
+export const createHeader = i18n => {
+  const Header = `
+    type Header {
+      # ${i18n.t`Header insulation nominal RSI (R-value Systeme International)`}
+      insulationNominalRsi: I18NFloat
+      # ${i18n.t`Header insulation nominal R-value`}
+      insulationNominalR: I18NFloat
+      # ${i18n.t`Header insulation effective RSI (R-value Systeme International)`}
+      insulationEffectiveRsi: I18NFloat
+      # ${i18n.t`Header insulation effective R-value`}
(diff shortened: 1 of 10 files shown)


api/src/schema/types/I18NBoolean.js
```diff
@@ -0,0 +1,9 @@
+import { GraphQLBoolean } from 'graphql'
+
+export const createI18NBoolean = i18n => {
+  const I18NBoolean = Object.create(GraphQLBoolean)
+  I18NBoolean.description = i18n.t`
+    The 'Boolean' scalar type represents 'true' or 'false'.
+  `
+  return I18NBoolean
+}
```
pcraig3: Is this just because it's cleaner syntax, or because it's a convention I wasn't following?

api/src/schema/Header.js
```diff
@@ -0,0 +1,27 @@
+export const createHeader = i18n => {
+  const Header = `
+    type Header {
```
pcraig3: It would be good to add the cache control stuff here as well.

`type Header @cacheControl(maxAge: 90) {`

sleepycat: Good idea.

api/src/schema/types/I18NBoolean.js
sleepycat: This is so these things can be easily pulled into multiple files. It's a little cleaner maybe but I don't love this either. I thinking is that the commits that follow will benefit from this. We'll see...

PR #277, opened 2018-02-28
Unzip in endpoint

sastels:
The Endpoint receives a zipped file. It unzips the file and sends the files inside up to Azure. This will allow rows to be updated (by overwriting the corresponding outdated Azure file)


extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -2,11 +2,11 @@
 import base64
 import typing
 from http import HTTPStatus
+import zipfile
 import flask
 from werkzeug import utils
 from azure.common import AzureMissingResourceHttpError
-from extract_endpoint import azure_utils
-from extract_endpoint import crypt_utils
+from extract_endpoint import azure_utils, crypt_utils
```
buckley-w-david: I don't particularly mind either way, but I know that Jason prefers imports like this be on 2 separate lines.

extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -59,25 +59,26 @@ def upload_file() -> typing.Tuple[str, int]:
...
+                                        data=base64.b64encode(file.read()).decode('utf-8'))
     if flask.request.form['signature'] != signature:
         flask.abort(HTTPStatus.BAD_REQUEST)
 
-    filename = utils.secure_filename(flask.request.form['filename'])
-
-    if not azure_utils.upload_bytes_to_azure(App.config['AZURE_COORDINATES'], file_contents, filename):
-        flask.abort(HTTPStatus.BAD_GATEWAY)
+    try:
+        file_z = zipfile.ZipFile(file)
```
buckley-w-david: Presumably my fears will be laid to rest once I read through the tests, but hasn't the file position been advanced to the end of the file by the read on `69`?

extract_endpoint/src/extract_endpoint/endpoint.py
sastels: yeah, I just realized that while mucking around in etl.

extract_endpoint/src/extract_endpoint/endpoint.py
sastels: yeah, I wondered about that too, but it seems to work anyway. However, it wouldn't hurt to throw a seek(0) in there just in case this behaviour changes (or is different on other systems).

extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -59,25 +60,27 @@ def upload_file() -> typing.Tuple[str, int]:
...
+    except zipfile.BadZipFile:
+        return "Bad Zipfile", HTTPStatus.BAD_REQUEST
+
+    for json_file in [file_z.open(zipinfo) for zipinfo in file_z.infolist()]:
+        if not azure_utils.upload_bytes_to_azure(App.config['AZURE_COORDINATES'], json_file.read(),
+                                                 utils.secure_filename(json_file.name)):
+            flask.abort(HTTPStatus.BAD_GATEWAY)
```
buckley-w-david: TIL what BAD_GATEWAY actually means

PR #281, opened 2018-02-28
Add option for transform to read from Azure

sastels:
Add a `--azure` flag that will use the contents of our Azure container instead of a local file.


etl/tests/conftest.py
```diff
@@ -81,6 +83,50 @@ def energuide_zip_fixture(tmpdir: py._path.local.LocalPath, energuide_fixture: s
...
+
+
+@pytest.fixture
+def azure_container() -> str:
+    return 'test-container'
+
+
+@pytest.fixture
+def azure_service(azure_container: str) -> blob.BlockBlobService:
+    reader.EXTRACT_ENDPOINT_STORAGE_ACCOUNT = 'devstoreaccount1'
```
JasonMWhite: @sastels @evadb You can't do this - this leads to flaky tests since this side effect lasts beyond the current test instantiation. Any later test that depends in any way on `reader.EXTRACT_ENDPOINT_STORAGE_ACCOUNT` will have this symbol set to this value.

Few things to take away from this:
1. use `monkeypatch.setattr` to do this exactly. The monkeypatch fixture will undo all the changes for you at the end of the test in teardown.
2. use `monkeypatch.setenv` to set the environment variables themselves (and unset them in teardown) to exercise that code itself
3. this is a code smell that should cause you to think that perhaps a refactor is required

I'm already working on a refactor that hits this, so I'll take care of it. Just FYI for next time.

evadb: Ahhh that makes sense, thanks Jason!

sastels: yeah, very good point 💯 . Other tests failing (or worse, passing) because of the side effects of a different test would be really miserable to deal with.

PR #286, opened 2018-03-01
Optional wall percentage

buckley-w-david:
This PR adds functionality to account for the fact that basement wall sections elements can be missing the percentage attribute, and if they are they are assumed to use 'the rest' of the total wall percentage.

So if there were 2 wall sections, one of which was 25%, the other did not have a percentage listed, it would be 75%


etl/src/energuide/embedded/basement.py
```diff
@@ -224,18 +226,35 @@ class BasementWall(_BasementWall):
...
+                   wall_perimeter: float,
+                   wall_height: float,
+                   tag: WallType,
+                   backup_percentage: float) -> 'BasementWall':
+
+        maybe_percentage = wall.attrib.get('percentage')
+        if maybe_percentage is None:
+            percentage = backup_percentage
+        else:
+            percentage = float(maybe_percentage)
```
sastels: could do this if/else in one line,
```
percentage = float(maybe_percentage) if maybe_percentage else backup_percentage
```
`¯\_(ツ)_/¯`

etl/src/energuide/embedded/basement.py
```diff
@@ -224,18 +226,35 @@ class BasementWall(_BasementWall):
...
+
         return BasementWall(
             wall_type=tag,
-            nominal_insulation=insulation.Insulation(float(wall.attrib['nominalRsi'])),
-            effective_insulation=insulation.Insulation(float(wall.attrib['rsi'])),
+            nominal_insulation=insulation.Insulation(nominal_insulation),
+            effective_insulation=insulation.Insulation(effective_insulation),
             composite_percentage=percentage,
-            wall_area=area.Area(wall_perimiter * wall_height * (percentage / 100))
+            wall_area=area.Area(wall_perimeter * wall_height * (percentage / 100))
```
sastels: lol good catch!

etl/src/energuide/embedded/basement.py
buckley-w-david: Already thought of that, would use backup if maybe was present but `0`, which while a weird situation is not what I want.

buckley-w-david: Of course I could just do an explicit check for `None` instead... I was tired when I submitted this PR

PR #287, opened 2018-03-01
Update deploy

szinck1:
This PR fixes formatting in circleci's config and tags the docker images with the last commit SHA instead of a version number.


sastels:
LGTM. @sleepycat , do you have any opinion on using the gitsha instead of a version # for the api docker images?

PR #293, opened 2018-03-01
Flask wrapper for Transform/Load

sastels:
A flask wrapper for the TL code that can run in Azure and be triggered by a post.


etl/src/energuide/cli.py
```diff
@@ -4,10 +4,10 @@
 from energuide import database
 from energuide import transform
 from energuide import extractor
-from energuide import logging
+from energuide import logger
```
sastels: Flask was really upset that we had a `logging` module and insisted on crashing.

etl/src/energuide/flask_app.py
```diff
@@ -0,0 +1,81 @@
...
+    return 'Alive!'
+
+
+@App.route('/robots933456.txt', methods=['GET'])
+def robots() -> None:
+    flask.abort(HTTPStatus.NOT_FOUND)
+
+
+@App.route('/run_tl', methods=['POST'])
+def run_tl() -> typing.Tuple[str, int]:
```
evadb: This seems to double up code from the load function in cli.py. Possibly necessary, but just an observation?

etl/src/energuide/flask_app.py
```diff
@@ -0,0 +1,81 @@
...
+
+    salt = flask.request.form['salt']
+    signature = flask.request.form['signature']
+    hasher = hashlib.new('sha3_256')
+    hasher.update((salt + App.config['SECRET_KEY']).encode())
+    salt_signature = hasher.hexdigest()
+    if salt_signature != signature:
+        return 'bad signature', HTTPStatus.BAD_REQUEST
+
+    azure_coords = transform.AzureCoordinates.from_env()
```
evadb: On the subject of cli.py, may be worth adding some of the start/finish logger calls here from the load function there?

PR #294, opened 2018-03-01
remove unused fixture file

JasonMWhite:
I believe this fixture is a relic of a previous era - it doesn't seem to be used anymore. 🔥


buckley-w-david:
I think you're right about nothing using it, but that is also the case for `conftest.py::sample_fixture`, which should also face the 🔥

JasonMWhite:
Ooh, good catch.

PR #298, opened 2018-03-01
remove drop_fields

JasonMWhite:
Simplifying the extractor a bit, cleaning up a bit of technical debt here to make it easier to hack on.


JasonMWhite:
Yes, the `purge_unknown` field does this for us already here: https://github.com/cds-snc/nrcan_api/blob/master/etl/src/energuide/extractor.py#L66

And we're testing that this is in fact the case here: https://github.com/cds-snc/nrcan_api/blob/master/etl/tests/test_extractor.py#L118

PR #299, opened 2018-03-01
[WIP] Add percentage efficiency for water heating

buckley-w-david:
This PR adds capability to understand when efficiency values are in the form as a percentage.

It also cleans up some of the exception handling


buckley-w-david:
Before this can be merged, I think we'll need an addition (in this branch) from either @pcraig3 or @sleepycat that changes the API side.

sleepycat:
@buckley-w-david I'll tackle that today.

JasonMWhite:
Just checking - did you get a 👍  on this PR before merging? Also note that the title was never changed from `WIP` which is a signal to _not_ give this PR a thorough review.

buckley-w-david:
Good point, completely missed that since I was going back and forth with Mike so much it just slipped my mind that anyone from the etl crew had yet to take a look at it.

PR #303, opened 2018-03-01
Only deploy endpoint from master

buckley-w-david:
The PR changes the CI configuration to only deploy_endpoint on master, and only if extact_endpoint job succeeded


JasonMWhite:
This would still deploy the endpoint if just the `extract_endpoint` tests passed. I propose instead that we fan-in all stages so that we only deploy anything IIF all tests pass.

PR #305, opened 2018-03-02
extractor progress bar

JasonMWhite:
This PR adds a progress indicator to the extraction function using `tqdm`. This is a nice feature that helps avoid the "is this thing working" symptom.

This PR also simplifies the extraction functions a bit, so they don't pass generators to functions and return generators needlessly. Just with a bit of simple refactoring, we get precisely the same result but with less embedded blocks.

### DIFF
--- a/etl/src/energuide/extractor.py
+++ b/etl/src/energuide/extractor.py
@@ -4,6 +4,7 @@ import typing
 import sys
 import zipfile
 import cerberus
+from tqdm import tqdm
 from energuide import element
 from energuide import reader
 from energuide import snippets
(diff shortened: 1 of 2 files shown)


etl/src/energuide/extractor.py
```diff
@@ -55,7 +53,12 @@ def _read_csv(filepath: str) -> typing.Iterator[reader.InputData]:
     except OverflowError:
         csv.field_size_limit(_WINDOWS_LONG_SIZE)
 
+    total_lines = 0
     with open(filepath, 'r', encoding='utf-8', newline='') as file:
+        for _ in file:
+            total_lines += 1
```
sastels: Could replace this with `total_lines = sum(1 for _ in file)`

etl/src/energuide/extractor.py
```diff
@@ -55,7 +53,12 @@ def _read_csv(filepath: str) -> typing.Iterator[reader.InputData]:
...
+
+    with tqdm(open(filepath, 'r', encoding='utf-8', newline=''), total=total_lines, unit=' lines') as file:
```
sastels: nice

etl/src/energuide/extractor.py
JasonMWhite: I was trying to find a better way. This is 👍

PR #306, opened 2018-03-02
Add pytest fixture and example test

evadb:
This PR creates a fixture that attempts to check whether or not the Azure emulator Azurite is running. The fixture checks for the following:

* If the CIRCLECI environment variable exists, run the tests
* If it doesn't exist, try binding to the port Azurite uses (10000)
* If a port error is received (in use), assume it is Azurite & try to run the tests
* If no port error, skip the tests

To use the fixture, add @pytest.mark.usefixtures('is_azurite_running') before any test that requires the Azure emulator

### DIFF
--- a/etl/tests/conftest.py
+++ b/etl/tests/conftest.py
@@ -93,4 +93,4 @@ def is_azurite_running() -> None:
         return
     else:
         sock.close()
-        pytest.skip()
+        pytest.skip("Azure simulator not running, test skipped")
--- a/etl/tests/conftest.py
+++ b/etl/tests/conftest.py
@@ -91,5 +91,3 @@ def is_azurite_running() -> None:
     else:
         sock.close()
         pytest.skip()
-
-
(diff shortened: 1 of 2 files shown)


etl/tests/conftest.py
```diff
@@ -78,3 +79,18 @@ def energuide_zip_fixture(tmpdir: py._path.local.LocalPath, energuide_fixture: s
     data = extractor.extract_data(energuide_fixture)
     extractor.write_data(data, outfile)
     return outfile
+
+
+@pytest.fixture
+def is_azurite_running() -> None:
```
sastels: how about we call this something like "skip_if_azure_simulator_not_running"?

etl/tests/test_azure_skip.py
```diff
@@ -0,0 +1,6 @@
+import pytest
```
sastels: 🔥

PR #309, opened 2018-03-02
Add foundations to schema

sleepycat:
Finally the foundations!


pcraig3:
```

pcraig3:
I understand that foundations are different than pillars, but you've gotta start somewhere.

pcraig3:
Also, those enums >.<

PR #313, opened 2018-03-02
:fire: reader, split transform from run

JasonMWhite:
This PR :fire: the `reader.py` and pulls the necessary functions directly into `transform.py`. It also starts to refactor the `transform.run` function by splitting the `transform`ing from the `load`ing to Mongo. A later refactor will :fire: `transform.run` entirely by putting a dedicated function for loading into a new module.


JasonMWhite:
Rebased to fix a merge conflict.

PR #314, opened 2018-03-02
:fire: transform.run

JasonMWhite:
This PR :fire: `transform.run` since it just calls `database.load` anyway. The "transform" and "load" functions are now separated nicely which simplifies tests.


JasonMWhite:
Rebased to address a merge conflict. Ready for review.

PR #320, opened 2018-03-05
Refactor post to endpoint tests

evadb:
This PR refactors the post_to_endpoint tests to use the CLI instead of directly calling the post_stream function! This lets us test that the CLI works properly and will error out when the right options aren't provided. 

In order to test with the CLI a zip file test fixture was created. Also, I removed leftover filename references from post_to_endpoint since I was there anyway.


extract_endpoint/tests/conftest.py
```diff
@@ -53,3 +54,15 @@ def sample_zipfile(sample_filenames: typing.Tuple[str, str],
     file_z.close()
     file.seek(0)
     return file
+
+
+@pytest.fixture
+def sample_zipfile_fixture(tmpdir: py._path.local.LocalPath,
```
sastels: This is almost the same as the fixture just above (`sample_zipfile`) - you could you just write that fixture to disk here.

PR #322, opened 2018-03-05
Add enum functions for foundation floors

sleepycat:


api/src/schema/index.js
```diff
@@ -553,6 +553,34 @@ const Schema = i18n => {
       foundationHeaderHeightMetres
       # ${i18n.t`Filter results for dwellings with a matching header height in feet (ft)`}
       foundationHeaderHeightFeet
+      # ${i18n.t`Filter for dwellings with a matching type of foundation floor type (en)`}
```
pcraig3: Hey, I've used the word 'specific' everywhere else, so do you think it would be like-for-like to just reword this, or do you think this fits better?

Personally, I think that "matching" sounds a bit more like equality, which is only one possible way of searching.

api/src/schema/index.js
```diff
@@ -553,6 +553,34 @@ const Schema = i18n => {
...
+      foundationFloorInsulationNominalRsi
+      # ${i18n.t`Filter for dwellings with a matching insulation nominal R-value on the foundation floor`}
+      foundationFloorInsulationNominalR
+      # ${i18n.t`Filter for dwellings with a matching insulation effective RSI (R-value Systeme International) for the foundation floor`}
+      foundationFloorInsulationEffectiveRsi
+      # ${i18n.t`Filter for dwellings with a matching insulation effective R-value for the foundation floor`}
+      foundationFloorInsulationEffectiveR
+      # ${i18n.t`Filter for dwellings where the foundation floor has a matching area in square metres (m2)`}
+      foundationFloorAreaMetres
+      # ${i18n.t`Filter for dwellings where the foundation floor has a matching area in square feet (ft)`}
```
pcraig3: Square feet has an `(ft2)` 🔍

api/test/enum.test.js
```diff
@@ -138,6 +138,21 @@ describe('Enum values', () => {
...
+    foundationFloorPerimeterFeet: { testValue: 113.998035144 },
+    // TODO: Uncomment and add appropriate values when issue #315 is resloved
+    // foundationFloorInsulationNominalRsi: { testValue: null },
+    // foundationFloorInsulationNominalR: { testValue: null },
+    // foundationFloorInsulationEffectiveRsi: { testValue: null },
+    // foundationFloorInsulationEffectiveR: { testValue: null },
+    // foundationFloorWidthMetres: { testValue: null },
+    // foundationFloorWidthFeet: { testValue: null },
+    // foundationFloorLengthMetres: { testValue: null },
+    // foundationFloorLengthFeet: { testValue: null },
```
pcraig3: Oh no! 🤒 

Is this going to be resolved soon (in the next day or two), or are we expecting this to be here for a while?

sleepycat: Yeah, there are null values in the test data. When we get the rest of the data I suspect we will find actual values. The goal of these tests is to show that the Mongo query object returned by these functions is capable of returning a value from the database (it's structure lines up with the structure of the document itself). All the `foundationFloor*` functions use the same query object, so if one of the tests pass we have proof that things are OK, but for completeness it'd be good to get these fixed once the data is available... whenever that is.

api/src/schema/index.js
sleepycat: :+1:

api/src/schema/index.js
sleepycat: :dart:

api/test/enum.test.js
pcraig3: Okay, so we can't currently search on these values because we can't match on nulls.

pcraig3:
👍

PR #326, opened 2018-03-06
Refactor azure/local reading

JasonMWhite:
This PR cleans up some of the local/Azure switching to use some duck-typing to abstract away the problem of reading extracted rows. That had the benefit of cleaning up tests considerably.

### DIFF
--- a/etl/tests/conftest.py
+++ b/etl/tests/conftest.py
@@ -10,6 +10,7 @@ import pytest
 from azure.storage import blob
 from energuide import database
 from energuide import extractor
+from energuide import transform
 
 
 def azure_emulator_is_running() -> bool:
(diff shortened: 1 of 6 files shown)


etl/tests/conftest.py
```diff
@@ -102,41 +103,45 @@ def sample_fixture() -> str:
...
+
 
-    azure_service = blob.BlockBlobService(account_name=account,
-                                          account_key=key,
-                                          custom_domain=domain)
-    azure_service.create_container(azure_container_name)
-    yield azure_service
-    azure_service.delete_container(azure_container_name)
+@pytest.fixture
+def azure_service(azure_coordinates: transform.AzureCoordinates) -> typing.Iterator[transform.AzureCoordinates]:
```
sastels: It's a bit confusing that `azure_service` returns `AzureCoordinates` rather than a `BlockBlobService`. Could this be renamed? even something like `azure_emulator` maybe?

etl/tests/test_transform.py
```diff
@@ -5,24 +5,50 @@
...
+    output = list(local_reader.extracted_rows())
+    unique_builders = {row['BUILDER'] for row in output}
+    assert len(output) == 14
+    assert len(unique_builders) == 14
 
-def test_transform_no_azure_no_filename() -> None:
-    with pytest.raises(ValueError):
-        list(transform.transform(False, None))
 
+def test_azure_reader(azure_reader: transform.AzureExtractReader) -> None:
```
sastels: really nice to have the readers separated so we can test them like this.

PR #327, opened 2018-03-06
Deployetl

szinck1:
Deploy the ETL to Docker Hub, tag images with SHA1 instead of $VERSION.


.circleci/config.yml
```diff
@@ -1,92 +1,87 @@
...
-      -
-        attach_workspace:
+      - attach_workspace: 
           at: /tmp/workspace
       - setup_remote_docker
-      -
-        run:
+      - run: 
           command: |
               mkdir ~/repo/api/dist
```
JasonMWhite: This whole section might be more palatable if we put it in a bash file in `etl/scripts`?

.circleci/config.yml
```diff
@@ -1,92 +1,87 @@
-version: 2
-jobs:
-  deploy_api:
-    docker:
-      -
+jobs: 
+  deploy_api:
```
JasonMWhite: Do we need 3 separate deployment jobs? Could we do it with 1 that deploys all three sections? Would use less containers on CI, of which we only have 4.

.circleci/config.yml
```diff
@@ -1,92 +1,87 @@
...
       - setup_remote_docker
-      -
-        run:
+      - run:
           command: |
               docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
-              docker build --build-arg VERSION="${VERSION}" -t "${DOCKER_REGISTRY}"/"${NAMESPACE}"/"nrcan_endpoint":"${VERSION}" -t "${DOCKER_REGISTRY}"/"${NAMESPACE}"/"nrcan_endpoint":"latest" .
-              docker push "${DOCKER_REGISTRY}"/"${NAMESPACE}"/"nrcan_endpoint":"${VERSION}"
-              docker push "${DOCKER_REGISTRY}"/"${NAMESPACE}"/"nrcan_endpoint":"latest"
+              docker build -t "${DOCKER_REGISTRY}/${NAMESPACE}/nrcan_etl:${CIRCLE_SHA1}" -t "${DOCKER_REGISTRY}/${NAMESPACE}/nrcan_etl:latest" .
```
JasonMWhite: These are all so similar, it would seem they could all be a single bash file with different arguments passed in.

.circleci/config.yml
```diff
@@ -1,92 +1,87 @@
...
               . ~/.venv/bin/activate
               pip install -r requirements.txt
               pip install -e .
           name: "Installing dependencies"
-      -
-        save_cache:
+      - save_cache: 
           key: "deps1-{{ .Branch }}-{{ checksum \"requirements.txt\" }}"
-          paths:
+          paths:
```
JasonMWhite: extra whitespace

.circleci/config.yml
```diff
@@ -100,255 +95,215 @@ jobs:
...
               . ~/.venv/bin/activate
               pip install -r requirements.txt
               pip install -e .
           name: "Installing dependencies"
-      -
-        save_cache:
+      - save_cache: 
           key: "deps2-{{ .Branch }}-{{ checksum \"requirements.txt\" }}"
-          paths:
+          paths:
```
JasonMWhite: whitespace

.circleci/config.yml
```diff
@@ -100,255 +95,215 @@ jobs:
...
+  extract_endpoint: 
+    docker: 
+      - image: "circleci/python:3.6.1"
+      - environment: executable=blob
         image: arafato/azurite
-    steps:
-      -
-        checkout:
+    steps: 
+      - checkout:
```
JasonMWhite: I appreciate the reformatting, but generally it would be better to do that in a separate, prior PR. Makes it easier to identify the critical new parts from the lower-risk reformatting.

.circleci/config.yml
```diff
@@ -100,255 +95,215 @@ jobs:
...
+      - run: 
           command: ~/repo/scripts/notify_failure.sh
           name: "Report failing Master"
           when: on_fail
     working_directory: ~/repo/api
-workflows:
-  version: 2
-  node_and_python:
-    jobs:
+version: 2
```
JasonMWhite: Why move this down?

.circleci/config.yml
```diff
@@ -100,255 +95,215 @@ jobs:
...
+workflows: 
+  node_and_python: 
+    jobs:
```
JasonMWhite: whitespace

szinck1:
I think I've addressed your concerns @JasonMWhite except for moving the deployment jobs into one big job. We probably could but I thought having them set up individually might give us more freedom to adjust parameters later on.  In any case, I wanted to address the other issues right away.

JasonMWhite:
Looks like the ETL tests aren't able to connect to MongoDB - connection refused.

buckley-w-david:
CI hasn't run on this PR for some reason

JasonMWhite (approved):

buckley-w-david (changes requested):
Not approved until CI has run and succeeded

buckley-w-david (approved):

PR #328, opened 2018-03-06
🔥 azure references in test_endpoint

evadb:
Removed timestamp related tests because those are now covered by post_to_endpoint with CLI refactor.

Burned references to azure fixtures in test_endpoint, since none of the tests are hitting (or need to hit) the azure storage.

### DIFF
--- a/extract_endpoint/tests/test_endpoint.py
+++ b/extract_endpoint/tests/test_endpoint.py
@@ -103,37 +103,6 @@ def check_file_in_azure(azure_service: blob.BlockBlobService,
     assert actual_blob.content == contents
 
 
-def test_upload_with_timestamp(azure_emulator_coords: azure_utils.StorageCoordinates,
-                               test_client: testing.FlaskClient,
-                               azure_service: blob.BlockBlobService,
-                               sample_timestamp: str,
-                               sample_salt: str,
-                               sample_zipfile_signature: str,
-                               sample_file_contents: str,
-                               sample_filenames: str,
-                               sample_zipfile: io.BytesIO) -> None:
-
-    post_return = test_client.post('/upload_file', data=dict(salt=sample_salt, signature=sample_zipfile_signature,
-                                                             timestamp=sample_timestamp,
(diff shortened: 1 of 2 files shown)


extract_endpoint/tests/test_endpoint.py
```diff
@@ -93,47 +93,6 @@ def test_timestamp_no_file(test_client: testing.FlaskClient) -> None:
...
-
-@pytest.mark.usefixtures('azure_service')
-def test_upload_without_timestamp(test_client: testing.FlaskClient,
-                                  sample_salt: str,
-                                  sample_zipfile_signature: str,
-                                  sample_zipfile: io.BytesIO) -> None:
-
-    post_return = test_client.post('/upload_file', data=dict(salt=sample_salt, signature=sample_zipfile_signature,
-                                                             file=(sample_zipfile, 'zipfile')))
-    assert post_return.status_code == HTTPStatus.BAD_REQUEST
```
sastels: I think we still need these two tests - otherwise `pytest tests/test_endpoint.py` will not do a full test of `endpoint.py`. I know we essentially duplicate this testing in `tests/test_post_endpoint.py` but I don't think it's good to rely on that file to ensure that `endpoint.py` is tested. Thoughts?

extract_endpoint/tests/test_endpoint.py
```diff
@@ -93,47 +93,6 @@ def test_timestamp_no_file(test_client: testing.FlaskClient) -> None:
...
-
-
-@pytest.mark.usefixtures('azure_service')
```
sastels: 👍 good catch noticing that we didn't need `azure_service` in all these tests.

extract_endpoint/tests/test_endpoint.py
evadb: Good catch! I agree and have put them back in, which does mean one azure reference needs to sneak back in 😭 I think there's value in refactoring these eventually so that the endpoint tests don't rely on actually uploading to azure, since we test that functionality in azure_utils already, but for now this will have to stay!

PR #329, opened 2018-03-06
add progress bar to transform, clean up output

JasonMWhite:
This PR addresses some issues I had trying to run the `transform` command on the sample data:
- no progress bar
- very messy/long output being displayed on error messages

### DIFF
--- a/etl/src/energuide/transform.py
+++ b/etl/src/energuide/transform.py
@@ -3,6 +3,7 @@ import json
 import os
 import typing
 import zipfile
+from tqdm import tqdm
 import typing_extensions
 from azure.storage import blob
 from energuide import dwelling
@@ -36,6 +37,9 @@ class ExtractProtocol(typing_extensions.Protocol):
     def extracted_rows(self) -> typing.Iterator[typing.Dict[str, typing.Any]]:
         pass
 
+    def num_rows(self) -> int:
+        pass
+
 
 class LocalExtractReader:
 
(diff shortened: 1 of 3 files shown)


etl/src/energuide/transform.py
```diff
@@ -47,28 +51,44 @@ def extracted_rows(self) -> typing.Iterator[typing.Dict[str, typing.Any]]:
...
                         if 'timestamp' not in blob_.name])
-        for file in files:
-            content = azure_service.get_blob_to_bytes(self._coords.container, file).content
+        return files
+
+    def extracted_rows(self) -> typing.Iterator[typing.Dict[str, typing.Any]]:
+        for file in self._files():
+            content = self._azure_service.get_blob_to_bytes(self._coords.container, file).content
             house = json.loads(content)
-            house['jsonFileName'] = content
```
JasonMWhite: This was the change that made the most difference to the messy output. On some errors, it was printing out the whole content instead of the filename.

etl/src/energuide/transform.py
```diff
@@ -89,8 +109,10 @@ def _generate_dwellings(grouped: typing.List[typing.Dict[str, typing.Any]]) -> t
     return None
 
 
-def transform(extract_reader: ExtractProtocol) -> typing.Iterator[dwelling.Dwelling]:
-    for row_group in _read_groups(extract_reader.extracted_rows()):
+def transform(extract_reader: ExtractProtocol, show_progress: bool = False) -> typing.Iterator[dwelling.Dwelling]:
```
JasonMWhite: Also, all tests that iterated through the `tqdm`-ified iterator caused console output, particularly when the tests failed. That made for some messy logs to parse through, and in one test where we were explicitly testing that a specific error message was logged, it caused that test to fail (since there was unrelated unexpected output to stderr).

etl/src/energuide/transform.py
sastels: 👍

PR #340, opened 2018-03-08
azure_utils refactor

evadb:
This PR refactors azure_utils to create an AzureStorage class rather than just pure functions. It also adds a LocalStorage class than can be used in the place of AzureStorage for testing, once additional refactoring is done.


extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,66 @@ class EnvVariables(enum.Enum):
     domain = 'EXTRACT_ENDPOINT_STORAGE_DOMAIN'
 
 
-def upload_bytes_to_azure(coords: StorageCoordinates, data: bytes, filename: str) -> bool:
-    azure_service = blob.BlockBlobService(account_name=coords.account,
-                                          account_key=coords.key,
-                                          custom_domain=coords.domain)
-    azure_service.create_blob_from_bytes(coords.container, filename, data)
-    return filename in [blob.name for blob in azure_service.list_blobs(coords.container)]
+class StorageProtocol(typing_extensions.Protocol):
```
sastels: maybe `write` and `read` would be better method names (ie more applicable to both cloud and local)

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,66 @@ class EnvVariables(enum.Enum):
...
+    def upload(self, data: bytes, filename: str) -> bool:
+        pass
 
+    def download(self, filename: str) -> str:
```
sastels: This should probably return bytes

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,66 @@ class EnvVariables(enum.Enum):
...
+
+    @property
+    def upload_run_count(self) -> int:
+        return self._upload_run_count
+
+    @property
+    def download_run_count(self) -> int:
+        return self._download_run_count
+
+    def upload(self, data: bytes, filename: str) -> bool:
```
sastels: you don't seem to be reading or writing from/to disk...

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,66 @@ class EnvVariables(enum.Enum):
...
+
+class AzureStorage:
+    def __init__(self, coords: StorageCoordinates) -> None:
+        self._coords = coords
+        self._azure: typing.Optional[blob.BlockBlobService] = None
+
+    @property
+    def _azure_service(self) -> blob.BlockBlobService:
+        if self._azure is None:
+            self._azure = blob.BlockBlobService(account_name=self._coords.account,
```
sastels: why not just set `_azure` in `init()`?

extract_endpoint/src/extract_endpoint/azure_utils.py
evadb: LocalStorage should probably be renamed TestStorage to be more specific. The goal of it isn't to store or download anything, rather to allow us to test in our unit tests whether the function we expect to be called is & that it's called the right number of times

extract_endpoint/src/extract_endpoint/azure_utils.py
sastels: though if we just have a MockStorage as the alternative, upload and download are better 👍

sastels: though if we just have a MockStorage as the alternative, upload and download are better 👍

extract_endpoint/src/extract_endpoint/azure_utils.py
evadb: fixed this!

extract_endpoint/src/extract_endpoint/azure_utils.py
evadb: Also fixed as discussed!

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,58 @@ class EnvVariables(enum.Enum):
...
+
+    def upload(self, data: bytes, filename: str) -> bool:
+        if data and filename:
+            self._upload_run_count += 1
+        return True
+
+    def download(self, filename: str) -> bytes:
+        if filename:
+            self._download_run_count += 1
+        return filename.encode()
```
JasonMWhite: I see what you're doing here. Here's an alternative:
- `upload` stores the data into a local dictionary using `filename` as the key
- `download` returns the data you put at the key of `filename`

That way setting up data to be downloaded is just `mock_storage.upload('filename', data)`

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,58 @@ class EnvVariables(enum.Enum):
...
+
+    def upload(self, data: bytes, filename: str) -> bool:
+        self._azure_service.create_blob_from_bytes(self._coords.container, filename, data)
+        return filename in [blob.name for blob in self._azure_service.list_blobs(self._coords.container)]
+
+    def download(self, filename: str) -> bytes:
+        return self._azure_service.get_blob_to_bytes(self._coords.container, filename).content
+
+
+def upload_bytes_to_azure(storage: StorageProtocol, data: bytes, filename: str) -> bool:
```
JasonMWhite: I don't think these functions should exist

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -5,6 +5,28 @@
 from extract_endpoint import azure_utils
 
 
+@pytest.fixture
+def azure_storage(azure_emulator_coords: azure_utils.StorageCoordinates) -> azure_utils.AzureStorage:
+    storage = azure_utils.AzureStorage(azure_emulator_coords)
+    return storage
```
JasonMWhite: 1-line these

extract_endpoint/src/extract_endpoint/azure_utils.py
evadb: 👍 this is such a good call, thank you! Code & tests are much cleaner this way

extract_endpoint/src/extract_endpoint/azure_utils.py
evadb: 👍 🔥 all the unnecessary things

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,39 @@ class EnvVariables(enum.Enum):
...
+    def __init__(self) -> None:
+        self._data: typing.Dict[str, bytes] = {}
+
+    def upload(self, data: bytes, filename: str) -> bool:
+        self._data.update({filename: data})
+        return True
+
+    def download(self, filename: str) -> typing.Optional[bytes]:
+        if filename in self._data:
+            return self._data.get(filename)
```
buckley-w-david: Since you already check for existence, this can be `return self._data[filename]`, which allows you to change the return type to `bytes` (which is what it really is, it should never return `None`, just raises an exception sometimes)

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,39 @@ class EnvVariables(enum.Enum):
...
+        else:
+            raise ValueError('File does not exist')
```
buckley-w-david: Perhaps a `LookupError` would be a better fit than `ValueError`? I'll leave that to you if you want to change or not, not a big deal and kind of up to personal preference.

extract_endpoint/src/extract_endpoint/azure_utils.py
evadb: Good call!

extract_endpoint/src/extract_endpoint/azure_utils.py
evadb: 👍

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,39 @@ class EnvVariables(enum.Enum):
...
+
+class MockStorage:
+    def __init__(self) -> None:
+        self._data: typing.Dict[str, bytes] = {}
+
+    def upload(self, data: bytes, filename: str) -> bool:
+        self._data.update({filename: data})
+        return True
+
+    def download(self, filename: str) -> typing.Optional[bytes]:
```
buckley-w-david: This return type can be just `bytes` instead of `typing.Optional[bytes]`

evadb: I could have sworn I changed this, apparently it sure helps to hit save before committing 😓

buckley-w-david: I hate it when I do that

extract_endpoint/src/extract_endpoint/azure_utils.py
```diff
@@ -17,16 +18,39 @@ class EnvVariables(enum.Enum):
...
+    def download(self, filename: str) -> bytes:
+        pass
 
-def download_bytes_from_azure(coords: StorageCoordinates, filename: str) -> str:
-    azure_service = blob.BlockBlobService(account_name=coords.account,
-                                          account_key=coords.key,
-                                          custom_domain=coords.domain)
-    return azure_service.get_blob_to_bytes(coords.container, filename).content
+
+class MockStorage:
```
buckley-w-david: As per IRL discussion, since this is code only for tests, it should live in a test file.

evadb: 👍

sastels:
😢

evadb:
haha I was sad too

moment o silence for my contributions to this project

PR #341, opened 2018-03-08
EVAL_ID in json filenames

buckley-w-david:
This PR adds the EVAL_ID to the json filenames in the zipfile produced by extract so that the json files can be sorted by filename prior to grouping.


buckley-w-david:
👍 for less wrong

PR #342, opened 2018-03-09
Robustify TL web app

sastels:
* when TL is triggered, respond immediately to Endpoint, and start running TL in a new thread
* if TL is triggered again before it has finished running, ignore the trigger request (and return a `TOO_MANY_REQUESTS` code

### DIFF
--- a/etl/tests/test_flask_app.py
+++ b/etl/tests/test_flask_app.py
@@ -44,15 +44,16 @@ def thread_runner() -> typing.Generator:
 
 
 def test_threadrunner(thread_runner: flask_app.ThreadRunner) -> None:
+    stay_asleep = True
 
     def sleeper() -> None:
-        time.sleep(0.5)
+        while stay_asleep:
+            pass
 
     thread_runner.start_new_thread(sleeper)
     assert thread_runner.is_thread_running()
-    time.sleep(0.4)
-    assert thread_runner.is_thread_running()
-    time.sleep(0.2)
+    stay_asleep = False
+    time.sleep(0.1)
     assert not thread_runner.is_thread_running()
 
 
--- a/etl/tests/test_flask_app.py
+++ b/etl/tests/test_flask_app.py
@@ -1,6 +1,7 @@
 from http import HTTPStatus
 import hashlib
 import time
+import typing
 import _pytest
 import pymongo
 import pytest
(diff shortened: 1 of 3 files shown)


etl/tests/test_flask_app.py
```diff
@@ -34,6 +36,26 @@ def test_client(monkeypatch: _pytest.monkeypatch.MonkeyPatch, database_name: str
...
 
 
+@pytest.fixture()
+def thread_runner() -> typing.Generator:
+    yield flask_app.ThreadRunner
+    while flask_app.ThreadRunner.is_thread_running():
+        time.sleep(0.1)
+
+
+def test_threadrunner(thread_runner: flask_app.ThreadRunner) -> None:
```
buckley-w-david: This seems like a flakey test.
Sleep is not guaranteed to block for exactly the time it is given, it is guaranteed to block for <i>at least</i> that long, and may run for an arbitrary amount longer due to scheduling of other activities.
I don't know of a great way around that, but maybe something like the `sleeper` function instead of just sleeping checking a `flag` variable that is scoped to `test_threadrunner` above it in a `while` loop, then switching that flag to turn it off?

etl/tests/test_flask_app.py
```diff
@@ -48,19 +70,34 @@ def test_robots(test_client: testing.FlaskClient) -> None:
...
     post_return = test_client.post('/run_tl', data=dict(salt=sample_salt, signature=sample_signature))
-    assert post_return.status_code == HTTPStatus.CREATED
+    assert post_return.status_code == HTTPStatus.OK
+    while thread_runner.is_thread_running():
+        time.sleep(0.1)
     assert mongo_client[database_name][collection].count() == 7
 
 
+@pytest.mark.timeout(10)
+@pytest.mark.usefixtures('populated_azure_emulator', 'thread_runner')
```
buckley-w-david: Why does this test need to use the `thread_runner` fixture?

sastels: To ensure that the Thread is torndown (ie waits until the thread is done) after the test finishes.

etl/tests/test_flask_app.py
```diff
@@ -34,6 +36,27 @@ def test_client(monkeypatch: _pytest.monkeypatch.MonkeyPatch, database_name: str
...
+    stay_asleep = True
+
+    def sleeper() -> None:
+        while stay_asleep:
+            pass
+
+    thread_runner.start_new_thread(sleeper)
+    assert thread_runner.is_thread_running()
+    stay_asleep = False
+    time.sleep(0.1)
```
buckley-w-david: I think instead of sleeping, it'd be better to explicitly wait until the thread is done. This can be done by `join`ing the thread. As it is the thread is hidden away as a private variable of ThreadRunner, but to get around that we could (and I think there is value in) add a `join` method to ThreadRunner that just calls join on the internal thread.

PR #344, opened 2018-03-09
Allow single pre-retrofit evaluations

buckley-w-david:
This PR allows 1 length groups of `ParsedDwellingDataRow` as long as they are pre-retrofit

### DIFF
--- a/etl/src/energuide/dwelling.py
+++ b/etl/src/energuide/dwelling.py
@@ -366,7 +366,10 @@ class Dwelling:
 
     @classmethod
     def _from_parsed_group(cls, data: typing.List[ParsedDwellingDataRow]) -> 'Dwelling':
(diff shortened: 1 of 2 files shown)


etl/src/energuide/dwelling.py
```diff
@@ -366,7 +366,10 @@ def __init__(self, *,
 
     @classmethod
     def _from_parsed_group(cls, data: typing.List[ParsedDwellingDataRow]) -> 'Dwelling':
-        if len(data) == 2:
+        if len(data) == 1 or len(data) == 2:
+            if len(data) == 1 and data[0].eval_type is EvaluationType.POST_RETROFIT:
```
sastels: might be cleaner to move this `if` up a level.

PR #351, opened 2018-03-12
Sort files in LocalExtractReader

buckley-w-david:
This PR changes LocalExtractReader to return `extracted_rows` in order sorted by the filename, which is assuming to be of the form `{EVAL_ID}_{BUILDER}`


etl/tests/test_transform.py
```diff
@@ -17,13 +18,24 @@ def azure_reader(populated_azure_emulator: transform.AzureCoordinates) -> transf
 
 def test_reader(local_reader: transform.LocalExtractReader) -> None:
     output = list(local_reader.extracted_rows())
+
     output = sorted(output, key=lambda row: row['BUILDER'])
     unique_builders = {row['BUILDER'] for row in output}
+
```
sastels: did you intend to add these two returns here?

etl/tests/test_transform.py
```diff
@@ -17,13 +18,24 @@ def azure_reader(populated_azure_emulator: transform.AzureCoordinates) -> transf
 
 def test_reader(local_reader: transform.LocalExtractReader) -> None:
     output = list(local_reader.extracted_rows())
+
     output = sorted(output, key=lambda row: row['BUILDER'])
```
sastels: you could combine these two `output` rows into one

etl/tests/test_transform.py
```diff
@@ -17,13 +18,24 @@ def azure_reader(populated_azure_emulator: transform.AzureCoordinates) -> transf
...
     assert len(output) == 14
     assert output[0]['BUILDER'] == '11W2D00606'
     assert len(unique_builders) == 14
 
 
+def test_reader_sorted_by_eval_id(local_reader: transform.LocalExtractReader) -> None:
+    output = list(local_reader.extracted_rows())
+
+    assert all(
+        typing.cast(str, current_row.get('EVAL_ID')) <= typing.cast(str, next_row.get('EVAL_ID'))
```
sastels: `current_row.get('EVAL_ID')` isn't already a string?

buckley-w-david: It is, but mypy doesn't know that. `typing.cast` doesn't actually do anything to do the data, it just returns exactly what you passed it, but it does inform the type system.

etl/tests/test_transform.py
buckley-w-david: Only kind of. Originally the test I added was just another assertion in this test, but I decided to factor it out into it's own dedicated test. When I was doing stuff in the other test I added some newlines to make it prettier/more readable to me.

sastels:
`yarn integration` fails locally now (if you first do a `make setup` as the docs instruct). I think this is because `make setup` uses `energuide --load` to populate mongo. I suspect this PR will fix that?

PR #354, opened 2018-03-12
Drain water heat recovery

buckley-w-david:
This PR adds an additional field for the efficiency of the Drain Water Heat Recovery system (if present).


buckley-w-david:
ping @sleepycat 
Looks like integration tests have survived

PR #355, opened 2018-03-13
Don't restart server if locale files change

pcraig3:
Adding `lingui compile` to the build step meant that the server kept restarting itself.
By ignoring the 'src/locale' folder, we can prevent this.


pcraig3:
Merging because this is a trivial change.

PR #356, opened 2018-03-13
Remove `I18N` prefix from our scalar types

pcraig3:
Changing all of the names of our overwritten types to their default names. This will make it easier for external clients to connect to our API, as having changed the names of all the primitive scalar types is possibly a weird gotcha.

`I18NDate` -> `Date`
`I18NBoolean` -> `Boolean`
`I18NInt` -> `Int`
`I18NString` -> `String`
`I18NFloat` -> `Float`

Also changed the return types of a couple fields in the final commit. @buckley-w-david, do those changes look okay to you?


buckley-w-david:
Those return type changes seem correct to me 👍

pcraig3:
Cool, I'll fix the linter error and then merge this in.

PR #360, opened 2018-03-13
Add API field documentation

pcraig3:
Since GraphQL can introspect itself (win!!), there are a few libraries that can quickly generate docs for us.

[`graphql-markdown`](https://www.npmjs.com/package/graphql-markdown) generates a markdown file listing the various types that are returned by the API.
 
😎👉  [how cool is this??](https://github.com/cds-snc/nrcan_api/blob/3f5d716414dc3936630b0a73ad115a929bf2edbd/api/docs.md) 👈 😎 

To generate the docs:
- the API must be running locally
- `yarn docs`

It's possible that the README stuff will get restructured as right now it's very "get the app started" focused, but for now this is a reasonable place for these docs to live.

***

Also, several of our users have asked us for the list of fields we'll be giving them, so once this is in we can link them to it.

### DIFF
--- a/api/docs.md
+++ b/api/docs.md
@@ -1,0 +1,3049 @@
+# Schema Types
+
+<details>
+  <summary><strong>Table of Contents</strong></summary>
+
+  * [Query](#query)
+  * [Objects](#objects)
+    * [Ceiling](#ceiling)
+    * [Door](#door)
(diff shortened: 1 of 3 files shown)


pcraig3:
Gonna merge this.

PR #362, opened 2018-03-13
[WIP] Triggers continued

evadb:
This PR follows the azure_utils refactor. It adds endpoint_triggers and endpoint_runner. The idea is to create a list of triggers to send to the runner & separate out a bit of the POST upload function (especially azure components). 

To start, this PR adds an upload to azure trigger and a mock version. 

This is turning out to be a larger refactor than I anticipated and I'm not sure of the value of a massive refactor right before the deadline but... I had fun doing it and this is cool, so it'll at least live in a WIP PR :D (until it gets 🗡 )

### DIFF
--- a/extract_endpoint/src/extract_endpoint/endpoint_runner.py
+++ b/extract_endpoint/src/extract_endpoint/endpoint_runner.py
@@ -1,15 +1,6 @@
 import typing
 from extract_endpoint.endpoint_triggers import EndpointTrigger
 
-# class EndpointRunner:
-#     def __init__(self, stream: typing.IO[bytes], timestamp: str, url: str) -> None:
-#         self.stream = stream
-#         self.timestamp = timestamp
-#         self.url = url
-#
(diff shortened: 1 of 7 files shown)


extract_endpoint/src/extract_endpoint/endpoint_runner.py
```diff
@@ -0,0 +1,13 @@
+import typing
+from extract_endpoint.endpoint_triggers import EndpointTrigger
+
+
+class EndpointRunner:
+    def __init__(self, triggers: typing.List[EndpointTrigger]) -> None:
+        self.triggers = triggers
+        return None
```
buckley-w-david: 🔥 Putting no return statement will result in a return value of `None`

evadb: Oh duh lol, what was past me doing. Thanks for the catch

PR #363, opened 2018-03-13
Fix Heating errors

buckley-w-david:
This PR introduces some changes to `Heating` to correctly parse Baseboard and ComboHeatDhw heater types.

ComboHeatDhw are parsed just like any other type, but Baseboards are structured differently, and fail the parsing, but are all configured the same way.

This PR also adds some missing energy source codes for different types of wood.

### DIFF
--- a/etl/src/energuide/embedded/heating.py
+++ b/etl/src/energuide/embedded/heating.py
@@ -134,7 +134,8 @@ class Heating(_Heating):
         try:
             steady_state_value = node.get('Type1/*/Specifications/@isSteadyState', str)
         except ElementGetValueError as exc:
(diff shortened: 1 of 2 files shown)


etl/src/energuide/embedded/heating.py
```diff
@@ -61,6 +62,8 @@ class Heating(_Heating):
         4: EnergySource.PROPANE,
         5: EnergySource.WOOD,
         6: EnergySource.WOOD,
+        7: EnergySource.WOOD,
+        8: EnergySource.WOOD,
```
pcraig3: ▶️ Glad to see we're moving four-wood with this. ▶️

buckley-w-david: :neutral_face:

PR #370, opened 2018-03-14
Cleanup connection string generation

buckley-w-david:
This PR changes how connection strings for MongoDB are generated, as well as adds a `production` flag to the CLI, indication what method of connection string generation should be used.

`--production` signals that it is an Atlas managed MongoDB instance, while `--local` is for anything else (ostensibly a local db, but not necessarily)

### DIFF
--- a/etl/src/energuide/cli.py
+++ b/etl/src/energuide/cli.py
@@ -42,7 +42,8 @@ def load(username: str,
          filename: typing.Optional[str],
          append: bool,
          progress: bool,
-         production: bool,) -> None:
+         production: bool,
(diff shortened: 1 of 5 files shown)


etl/src/energuide/cli.py
```diff
@@ -37,12 +41,15 @@ def load(username: str,
          azure: bool,
          filename: typing.Optional[str],
          append: bool,
-         progress: bool) -> None:
+         progress: bool,
+         production: bool,) -> None:
```
sastels: 🔥 the trailing comma

PR #371, opened 2018-03-14
Update in place

buckley-w-david:
This PR implements the changes suggested by @JasonMWhite 
Instead of dropping the collection and loading from scratch (resulting in some downtime), load now updates each record in place, creating new ones if needed.

### DIFF
--- a/etl/src/energuide/database.py
+++ b/etl/src/energuide/database.py
@@ -82,14 +82,16 @@ def load(coords: DatabaseCoordinates,
          database_name: str,
          collection_name: str,
          data: typing.Iterable[dwelling.Dwelling],
-         append: bool = False) -> None:
+         update: bool = True) -> None:
     client: pymongo.MongoClient
     with mongo_client(coords) as client:
         database = client[database_name]
         collection = database[collection_name]
 
-        if not append:
+        if not update:
             collection.drop()
 
-        for data_to_load in _chunk_data(data):
-            collection.insert_many(data_to_load)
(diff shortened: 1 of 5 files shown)


etl/tests/test_database.py
```diff
@@ -22,5 +22,41 @@ def test_load(database_coordinates: database.DatabaseCoordinates,
...
+    )
+
+    database.load(
+        coords=database_coordinates,
+        database_name=database_name,
+        collection_name=collection,
+        data=load_data,
+        update=True
+    )
+    assert mongo_client[database_name][collection].count() == 4
```
sastels: should we check that house_id 1  changed appropriately?

etl/src/energuide/database.py
```diff
@@ -82,14 +82,15 @@ def load(coords: DatabaseCoordinates,
...
 
-        if not append:
+        if not update:
             collection.drop()
 
-        for data_to_load in _chunk_data(data):
-            collection.insert_many(data_to_load)
+        for row in data:
+            data_row = row.to_dict()
+            collection.update({'houseId': data_row['houseId']}, data_row, upsert=True)
```
sastels: upsert?

buckley-w-david: If it doesn't exist, add it.

etl/tests/test_database.py
buckley-w-david: Probably, I'll add that.

PR #372, opened 2018-03-15
Added logging to endpoint

sastels:
Do some proper logging in the Endpoint. 

Ended up just copying over `energuide/logger.py` rather than mess around with installing energuide in endpoint.

### DIFF
--- a/extract_endpoint/src/extract_endpoint/logger.py
+++ b/extract_endpoint/src/extract_endpoint/logger.py
@@ -1,0 +1,25 @@
+import logging
+
+
+def unwrap_exception_message(exc: BaseException, join: str = ' - ') -> str:
+    if exc.__context__:
+        if exc.args:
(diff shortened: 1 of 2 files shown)


extract_endpoint/src/extract_endpoint/logger.py
```diff
@@ -0,0 +1,25 @@
...
+    if not logger.hasHandlers():
+        handler = logging.StreamHandler()
+        formatter = logging.Formatter(
+            fmt='%(asctime)s - %(levelname)s - %(name)s: %(message)s',
+            datefmt='%Y-%m-%dT%H:%M:%S%z'
+        )
+        handler.setFormatter(formatter)
+        logger.addHandler(handler)
+
+    return logger
```
buckley-w-david: THIEF!

evadb: 😨

PR #377, opened 2018-03-16
Add azure skip if azure isn't running

evadb:
It would probably be helpful to add in the automatic test skip from etl if azure emulator isn't running, for tests that require it. This is handy because if someone starts the tests without starting the emulator, the tests that require it will hang with no feedback.

🎉 
screen shot 2018-03-16 at 11 01 53 am

### DIFF
--- a/extract_endpoint/tests/test_azure_utils.py
+++ b/extract_endpoint/tests/test_azure_utils.py
@@ -31,6 +31,7 @@ def put_file_in_azure(azure_emulator_coords: azure_utils.StorageCoordinates,
     azure_service.delete_blob(azure_emulator_coords.container, filename)
 
 
+@pytest.mark.usefixtures('azure_service')
 def check_file_in_azure(azure_service: blob.BlockBlobService,
                         azure_emulator_coords: azure_utils.StorageCoordinates,
                         filename: str,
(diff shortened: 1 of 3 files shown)


extract_endpoint/tests/test_azure_utils.py
```diff
@@ -41,6 +42,7 @@ def check_file_in_azure(azure_service: blob.BlockBlobService,
     assert actual_blob.content == contents
 
 
+@pytest.mark.usefixtures('azure_service')
 def test_upload_bytes(azure_emulator_coords: azure_utils.StorageCoordinates,
                       azure_service: blob.BlockBlobService,
```
sastels: azure_service already in parameters

extract_endpoint/tests/test_azure_utils.py
```diff
@@ -31,6 +31,7 @@ def put_file_in_azure(azure_emulator_coords: azure_utils.StorageCoordinates,
     azure_service.delete_blob(azure_emulator_coords.container, filename)
 
 
+@pytest.mark.usefixtures('azure_service')
 def check_file_in_azure(azure_service: blob.BlockBlobService,
```
sastels: azure_service already in parameters

extract_endpoint/tests/test_azure_utils.py
evadb: Good catch! Unfortunately, without adding the mark.usefixtures, the skip doesn't work :(

sastels:
😭

PR #383, opened 2018-03-16
Compress data in zip file

buckley-w-david:
This PR adds an option to `write_data` that signals to compress the data in the output zip file.


etl/src/energuide/extractor.py
```diff
@@ -121,7 +121,7 @@ def extract_data(input_path: str,
 def write_data(data: typing.Iterable[typing.Optional[typing.Dict[str, typing.Any]]],
                output_path: str) -> typing.Tuple[int, int]:
     records_written, records_failed = 0, 0
-    with zipfile.ZipFile(output_path, mode='w') as output_zip:
+    with zipfile.ZipFile(output_path, mode='w', compression=zipfile.ZIP_DEFLATED) as output_zip:
```
sastels: Bad zipfile!

PR #384, opened 2018-03-19
do azure upload in a thread

sastels:
Local post sometimes hangs when doing a big upload to Endpoint, despite Endpoint and TL App working correctly. Solution: have Endpoint return `OK` when it receives and validates the post, and run the Azure upload on a new thread. This is similar to how the TL App works.

### DIFF
--- a/extract_endpoint/src/extract_endpoint/endpoint.py
+++ b/extract_endpoint/src/extract_endpoint/endpoint.py
@@ -1,5 +1,8 @@
 import os
+import io
+import threading
 import hashlib
+import functools
 import secrets
 import typing
 from http import HTTPStatus
@@ -31,6 +34,24 @@ App.config.update(dict(
         domain=os.environ.get(azure_utils.EnvVariables.domain.value, None)
     )
 ))
+
+
+class ThreadRunner:
+    _running_thread: typing.Optional[threading.Thread] = None
+
+    @classmethod
+    def start_new_thread(cls, target: typing.Callable[[], None]) -> None:
+        cls._running_thread = threading.Thread(target=target)
+        cls._running_thread.start()
+
+    @classmethod
+    def is_thread_running(cls) -> bool:
+        return cls._running_thread is not None and cls._running_thread.is_alive()
+
+    @classmethod
+    def join(cls) -> None:
+        if cls._running_thread is not None:
+            cls._running_thread.join()
 
 
 def _run_tl_url() -> str:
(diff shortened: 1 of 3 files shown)


extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -126,28 +158,23 @@ def upload_file() -> typing.Tuple[str, int]:
...
     if not azure_utils.upload_bytes_to_azure(App.config['AZURE_COORDINATES'], timestamp.encode(), TIMESTAMP_FILENAME):
         LOGGER.warning("Timestamp upload to Azure storage failed")
         flask.abort(HTTPStatus.BAD_GATEWAY)
 
-    run_tl_return_code = run_tl()
-    LOGGER.info(f"TL returned {run_tl_return_code}")
-    if run_tl_return_code in [HTTPStatus.OK, HTTPStatus.TOO_MANY_REQUESTS]:
-        return '', run_tl_return_code
-    return '', HTTPStatus.BAD_GATEWAY
+    ThreadRunner.start_new_thread(target=functools.partial(unzip_upload_run_tl, data=file.read()))
```
sastels: python closes files when they go to a thread - why? :(  As a work around, pass the file contents and recreate the Zipfile in the thread.

buckley-w-david: Annoying, but I guess that makes sense

extract_endpoint/src/extract_endpoint/endpoint.py
```diff
@@ -126,28 +157,23 @@ def upload_file() -> typing.Tuple[str, int]:
         flask.abort(HTTPStatus.BAD_REQUEST)
 
     try:
-        file_z = zipfile.ZipFile(file)
+        zipfile.ZipFile(file)
+        file.seek(0)
```
buckley-w-david: Are you seeking here because `zipfile.ZipFile` advances the file handle?

extract_endpoint/tests/test_endpoint.py
```diff
@@ -84,6 +84,36 @@ def mock_send_to_tl(data: typing.Dict[str, str]) -> int:
...
+    thread_runner.join()
+    assert not thread_runner.is_thread_running()
+
+
+def check_file_in_azure(azure_service: blob.BlockBlobService,
+                        azure_emulator_coords: azure_utils.StorageCoordinates,
+                        filename: str,
+                        contents: str) -> None:
+
+    assert filename in [blob.name for blob in azure_service.list_blobs(azure_emulator_coords.container)]
```
buckley-w-david: Probably doesn't matter in this context, but for future reference a `set` comprehension would be better since `in` checks are O(1)

extract_endpoint/src/extract_endpoint/endpoint.py
sastels: yes

extract_endpoint/tests/test_endpoint.py
sastels: yeah, hash tables ftw. In this case the list only contains 3 objects.

PR #387, opened 2018-03-20
[WIP] only run TL on new or related data

sastels:
Only run TL on Azure files that were created since the last time TL started running, or have the same `eval_id` as one of those files.


etl/src/energuide/transform.py
```diff
@@ -73,21 +75,45 @@ def _azure_service(self) -> blob.BlockBlobService:
...
+
+        for file in self._new_files:
             content = self._azure_service.get_blob_to_bytes(self._coords.container, file).content
             house = json.loads(content)
             house['jsonFileName'] = file
             yield house
 
     def num_rows(self) -> int:
-        return len(self._files())
+        if self._new_files is None:
```
buckley-w-david: Why not make `self._new_files` a property like `self._azure_service` so you don't need to duplicate these `if ... is None` checks.

etl/src/energuide/database.py
```diff
@@ -84,5 +84,3 @@ def load(coords: DatabaseCoordinates,
             existing.discard(row.house_id)
             data_row = row.to_dict()
             collection.update({'houseId': data_row['houseId']}, data_row, upsert=True)
-
-        collection.remove({'houseId': {'$in': list(existing)}})
```
buckley-w-david: If we're removing this, we can also remove the `existing` set entirely

etl/tests/test_transform.py
```diff
@@ -28,7 +30,7 @@ def test_reader_num_rows(local_reader: transform.LocalExtractReader) -> None:
     assert local_reader.num_rows() == 14
 
 
-def test_azure_reader(azure_reader: transform.AzureExtractReader) -> None:
+def test_azure_reader_extracted_rows(azure_reader: transform.AzureExtractReader,) -> None:
```
evadb: is there a reason for the extra comma?

sastels: good catch :)

etl/src/energuide/flask_app.py
```diff
@@ -57,9 +57,9 @@ def _run_tl_and_verify() -> None:
     with database.mongo_client(DATABASE_COORDS) as mongo_client:
         records_created = mongo_client[DATABASE_NAME][COLLECTION].count()
     if records_created == 0:
-        LOGGER.warning('Error, no records created')
+        LOGGER.warning('TL App Error, no records created')
     else:
-        LOGGER.info(f'Success, {records_created} created')
+        LOGGER.info(f'TL App success, {records_created} rows now in the database')
```
buckley-w-david: Don't know how necessary the `TL App` part of these messages is, it should display `energuide.flask_app` in the log message.

sastels: Unfortunately it just puts `__main__`

buckley-w-david: Well that sucks

PR #409, opened 2018-03-23
Walls remastered

buckley-w-david:
This PR changes walls from a specifically structured dictionary to a dedicated class

This allows for better type checking, and a cleaner implementation.


evadb:
is this the walls HD remastered edition?

PR #414, opened 2018-03-24
Add "F" to EvaluationType

buckley-w-david:
This PR adds "F" to EvaluationType, as while it is not still in use, it was in use for some time in the historical data that is in scope.


nrcan_etl/src/energuide/embedded/evaluation_type.py
```diff
@@ -5,12 +5,15 @@
 class EvaluationType(enum.Enum):
     PRE_RETROFIT = 'D'
     POST_RETROFIT = 'E'
+    INSENTIVE_PROGRAM = 'F'
```
evadb: Does "INSENTIVE" stand for something in particular? Is it possibly INCENTIVE?

buckley-w-david: Good catch, spelling is my weak point

nrcan_etl/src/energuide/embedded/evaluation_type.py
```diff
@@ -5,12 +5,15 @@
...
 
     @classmethod
     def from_code(cls, code: str) -> 'EvaluationType':
         if code == cls.PRE_RETROFIT.value:
             return EvaluationType.PRE_RETROFIT
         elif code == cls.POST_RETROFIT.value:
             return EvaluationType.POST_RETROFIT
+        elif code == cls.INSENTIVE_PROGRAM:
```
evadb: Should they all be changed to INCENTIVE?

nrcan_etl/src/energuide/embedded/evaluation_type.py
evadb: I became a programmer to avoid spelling tbh

PR #415, opened 2018-03-24
Docker ignore nrcan_etl

buckley-w-david:
If `etl` is in the .dockerignore file, then the new ETL `nrcan_etl` likely should be as well.


buckley-w-david:
After talking to @szinck1 last week, this shouldn't be needed.

PR #417, opened 2018-03-25
Remove unnecessary API fields 

evadb:
The scope of the project has been changed to a small set of fields pulled from the TSV, rather than the XML. The focus has also been narrowed to a set of priority fields that our partners would like to see in the api. 

This PR strips the none-priority fields from the api. Some fields may get readded this week, but in a drastically different form (the data from the TSV is much less granular than the XML). 

I haven't touched the lingui stuff yet. I have bad memories from trying to remove/add things and somehow breaking all the things. Will take a look at that in a separate PR. 

tl;dr 

screen shot 2018-03-26 at 10 04 12 am


pcraig3:
> unnecessary

PR #418, opened 2018-03-26
Add flask_app to nrcan_etl

evadb:
The PR copies and pastes the flask_app portion of the old energuide to the new energuide. Only change is the test now checks for 11 records instead of 7 since the testing data set changed. 

This is necessary for integration testing & api stuff


nrcan_etl/src/energuide/flask_app.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@App.route('/', methods=['GET'])
+def frontend() -> str:
+    return ''
+
+
+@App.route('/test_alive', methods=['GET'])
+def test_alive() -> str:
+    return 'Alive!'
```
sastels: You can 🔥 this, the '/' route can be used for testing if the app is up.

nrcan_etl/tests/test_flask_app.py
```diff
@@ -0,0 +1,156 @@
...
+    assert not thread_runner.is_thread_running()
+
+
+def test_frontend(test_client: testing.FlaskClient) -> None:
+    assert test_client.get('/').status_code == HTTPStatus.OK
+
+
+def test_alive(test_client: testing.FlaskClient) -> None:
+    get_return = test_client.get('/test_alive')
+    assert b'Alive' in get_return.data
```
sastels: and then 🔥 this

PR #431, opened 2018-03-27
Keep F evaluations

buckley-w-david:
This PR refactors `_filter_dummy_evaluations` to...
1. Not filter out `F` evaluations
2. Be prettier


buckley-w-david:
No, but that's only because none of our manually crafted test data have F evaluations. 

An argument can be made that we should change that though.

PR #434, opened 2018-03-27
Add ersRating

evadb:
The branch name lies - this PR adds ersRating back to the api, in new (improved? who knows) form. Also with bonus unit & integration tests, modelled off of @pcraig3 's Wall PR from yesterday.

ersRating looks like this now:

```
ersRating: {
            measurement: 
            upgrade: 
}
```

### DIFF
--- a/api/src/schema/index.js
+++ b/api/src/schema/index.js
@@ -105,7 +105,7 @@ const Schema = i18n => {
       # ${i18n.t`A list of upgrades that would improve energy efficiency`}
       energyUpgrades: [Upgrade]
       # ${i18n.t`The EnerGuide Rating calculated for this evaluation`}
-      ersRating: Int
+      ersRating: ErsRating
       walls: Wall
     }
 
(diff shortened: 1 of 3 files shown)


api/src/schema/index.js
```diff
@@ -125,6 +125,11 @@ const Schema = i18n => {
       evaluations: [Evaluation]
     }
 
+    type ErsRating @cacheControl(maxAge: 90) {
+      measurement: Int
+      upgrade: Int
+    }
+
```
pcraig3: Do we know that these are always integers? I used floats for the previous values because I was looking at numbers like `12.0`.

pcraig3: As a side note, it's a shame that `measurement` and `upgrade` return data in a different format than `walls`, which also has `measurement` and `upgrade` keys. Nothing we can do, I guess, but it might surprise consumers of the API down the road.

evadb: oh man you're totally right, good call, ty!

Also a good flag, I'll ask David about that

PR #436, opened 2018-03-28
Test dummy evaluations are filtered out

buckley-w-david:
This PR adds some a missing test to ensure that dwellings throw away "dummy" pre-evaluations (defined as pre-evaluations that have the same ENTRYDATE as another post-evaluation)

### DIFF
--- a/nrcan_etl/src/energuide/embedded/evaluation_type.py
+++ b/nrcan_etl/src/energuide/embedded/evaluation_type.py
@@ -14,6 +14,6 @@ class EvaluationType(enum.Enum):
             return EvaluationType.PRE_RETROFIT
         elif code == cls.POST_RETROFIT.value:
             return EvaluationType.POST_RETROFIT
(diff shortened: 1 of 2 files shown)


nrcan_etl/src/energuide/embedded/evaluation_type.py
```diff
@@ -14,6 +14,6 @@ def from_code(cls, code: str) -> 'EvaluationType':
             return EvaluationType.PRE_RETROFIT
         elif code == cls.POST_RETROFIT.value:
             return EvaluationType.POST_RETROFIT
-        elif code == cls.INCENTIVE_PROGRAM:
+        elif code == cls.INCENTIVE_PROGRAM.value:
```
sastels: whoa, shouldn't there have been a test that caught this?

buckley-w-david: This is the test that caught it.
I agree it should have existed before, but it was missed.

PR #439, opened 2018-03-28
Add EghRating

XavierPhanel:
Add eghRating & API


evadb:
@pcraig3 my bad, not sure how that snuck in there

pcraig3:
Looking at this again, I have realised something.

`ergRating` and `egsRating` are both the same internally. They both have two floats with the same keys. This means that  just come up with a general name for that data representation and then assign it to both.

So.

Instead of this 👎 

```
    type Evaluation @cacheControl(maxAge: 90) {
      ersRating: ErsRating
      eghRating: EghRating
    }

    type ErsRating @cacheControl(maxAge: 90) {
      measurement: Float
      upgrade: Float
    }
    type EghRating @cacheControl(maxAge: 90) {
      measurement: Float
      upgrade: Float
    }
```

We can do this 👍 
```
    type Evaluation @cacheControl(maxAge: 90) {
      ersRating: Rating
      eghRating: Rating
    }

    type Rating @cacheControl(maxAge: 90) {
      measurement: Float
      upgrade: Float
    }
```

pcraig3:
This is similar to how [`WallMeasurement` is re-used in the `Wall` type](https://github.com/cds-snc/nrcan_api/blob/08f2c15b00ef9e5cd2a5f47607256e3330801dc6/api/src/schema/index.js#L79-L82).

PR #440, opened 2018-03-28
adapt README and Makefile(s) to nrcan_etl

buckley-w-david:
This PR fixes some missing changes to the README referencing the old `etl`, and also changes the makefile(s) to now work with the new `nrcan_etl`


pcraig3:
What happens to the original `/etl` code? Are we keeping it as it is but not using it anymore?

buckley-w-david:
For now it's just been frozen as we transition stuff from it to the `/nrcan_etl`, nothing has been discussed as to what to do with the old code.

So so far keeping it as is but not using it.

pcraig3:
Okay, 👍 from me too then.

PR #443, opened 2018-03-28
Add house type to evaluation

evadb:
This PR adds houseType to Evaluation, which is a string that describes the type of house being evaluated. We decided to add this to Evaluation instead of Dwelling due to the fact that major renovations can lead to a changed house type between evaluations.


pcraig3:
Seems legit.

PR #444, opened 2018-03-28
Add heated floor area to Evaluation

evadb:
This PR refactors heatedFloorArea. The new heatedFloorArea is just a Float value in the evaluation.


pcraig3:
Does the float represent surface area?

pcraig3:
Okay, if it's surface area in meters squared, make sure that is written into the field description/documentation.

PR #446, opened 2018-03-28
Add greenhouseGasEmissions

evadb:
this PR adds the greenhouseGasEmissions field to the API. Very similar to previous fields and uses Rating type.

One more field to go


pcraig3:
👍 users were always asking for this

PR #447, opened 2018-03-28
Add energyIntensity field

evadb:
Saved the most intense field for last.

This PR adds energyIntensity as a field to the API. Same as the last ones, it uses the Rating type. This also marks the last of the priority api fields, yay 🎉


buckley-w-david:
🎉🎉🎉🎉

XavierPhanel:
LGTM

PR #454, opened 2018-04-03
bulk update instead of one at a time

buckley-w-david:
This PR changes the database loading code to chunk up the documents into bulk replace operations 1000 documents at a time.

### DIFF
--- a/nrcan_etl/src/energuide/database.py
+++ b/nrcan_etl/src/energuide/database.py
@@ -67,7 +67,7 @@ def mongo_client(database_coordinates: DatabaseCoordinates) -> typing.Iterable[p
 CHUNKSIZE = 1000
 
 T = typing.TypeVar('T')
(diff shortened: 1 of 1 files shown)


nrcan_etl/src/energuide/database.py
```diff
@@ -64,6 +64,23 @@ def mongo_client(database_coordinates: DatabaseCoordinates) -> typing.Iterable[p
         yield client
 
 
+CHUNKSIZE = 1000
+
+T = typing.TypeVar('T')
+def _chunk(data: typing.Iterable[T], size=CHUNKSIZE) -> typing.Iterable[typing.List[T]]:
```
sastels: should probably give a type to `size`

buckley-w-david: You are correct

PR #455, opened 2018-04-03
Add lingui translations for new fields

evadb:
This PR adds descriptions in french & english to the new API fields. WIP while waiting on translations


api/src/locale/en/messages.json
```diff
@@ -287,16 +336,61 @@
...
     "translation": "",
     "origin": [
       [
         "src/schema/index.js",
-        92
+        109
+      ]
+    ]
+  },
+  "The EnerGuide Rating calculated for this evaluation, actual rating": {
```
buckley-w-david: I should have mentioned before, this is a deprecated field, so newer evaluations will have this be NULL, we should probably get that across somehow

api/src/schema/index.js
```diff
@@ -64,18 +64,27 @@ const Schema = i18n => {
...
     type WallMeasurement @cacheControl(maxAge: 90) {
+      # ${i18n.t`Description of wall insulation`}
       insulation: [Insulation]
+      # ${i18n.t`Heat loss through walls in MJ`}
       heatLost: Float
     }
 
+    # ${i18n.t`Material used to insulate something`}
     type Insulation @cacheControl(maxAge: 90) {
+      # ${i18n.t`The proportion of total element that insulation makes up`}
```
buckley-w-david: "proportion of the total element" sounds odd, maybe it's just the "element" part.

Maybe "proportion of the total wall insulation that this element represents", or something similar"

evadb: It's a bit weird, but Insulation is a general type now so the description has to fit any element it might be used with

PR #456, opened 2018-04-03
Better region parsing

buckley-w-david:
This PR makes `region.Region.from_data` much more robust in the region strings it will be able to understand. This is necessary due to the fact that the HOUSEREGION TSV field seems to be a manual entry field, so there are typo's, shorthands, etc.

Multiple new interpretation paths have been added
1. French names are also checked
2. "Alternate" names have been added, these are names that one could reasonable expect a person to enter as the name of a region, and the same has been done with codes.
3. A prefix check is done if exact matches are not found, to see if a shorthand was used.
4. If all previous checks fail, a Levenshtein minimum number of edits calculation is preformed and if there is a match with above the minimum acceptable ratio, it is selected.

Along with those alternative interpretation paths, non alphabetic characters are now also stripped out of the string.


nrcan_etl/src/energuide/embedded/region.py
```diff
@@ -20,22 +57,68 @@ class Region(enum.Enum):
...
+                return Region[region]
+
+        # Code separated by _
+        code = data.replace('_', '')
+        region = cls._from_code(code)
+        if region:
+            return region
+
+        # Levenshtein minimum number of edits to closest region
+        ratios = [(Levenshtein.ratio(region, data), region) for region in Region.__members__.keys() - {'UNKNOWN'}]
```
sastels: wicked, I remember doing this stuff years ago! :)

nrcan_etl/src/energuide/embedded/region.py
```diff
@@ -20,22 +57,68 @@ class Region(enum.Enum):
...
+        ratio, region = max(ratios)
+
+        if ratio > _MINIMUM_SIMILARITY:
+            LOGGER.warning(f'SIMILARITY: Interpreting {data} as {region}')
```
sastels: should this be a warning or info? isn't this the expected behaviour if the data is clean?

nrcan_etl/src/energuide/embedded/region.py
```diff
@@ -20,22 +57,68 @@ class Region(enum.Enum):
...
+            return Region[region]
+
         return None
 
     @classmethod
     def from_data(cls, data: str) -> 'Region':
-        output = cls._from_name(data)
+        if len(data) < 2:
+            return Region.UNKNOWN
```
sastels: Wouldn't this case fall through and get `Region.UNKNOWN` on line 122?

nrcan_etl/tests/embedded/test_region.py
```diff
@@ -33,5 +91,16 @@ def test_from_code(self):
...
+            'NWT',
+        ]
+        output = [region.Region.from_data(row) for row in data]
+        assert output == [
+            region.Region.PRINCE_EDWARD_ISLAND,
+            region.Region.NORTHWEST_TERRITORIES,
+        ]
+
     def test_from_unknown_code(self):
         assert region.Region.from_data('CA') == region.Region.UNKNOWN
```
sastels: trailing newline

nrcan_etl/src/energuide/embedded/region.py
```diff
@@ -1,5 +1,42 @@
...
+    'YUKON_TERRITORY': 'YUKON',
+    'NORTHWEST_TERRITORY': 'NORTHWEST_TERRITORIES',
+}
+
+_ALTERNATIVE_CODES = {
+    'PEI': 'PE',
+    'NWT': 'NT'
+}
+
+_MINIMUM_SIMILARITY = 0.6
```
sastels: Why did you pick this value (just curious)?

nrcan_etl/src/energuide/embedded/region.py
buckley-w-david: This is a guard against the prefix matching in the fuzzy matching. I don't want a single "N" to be a valid prefix match, since it's ambiguous (Could be newfoundland, nova scotia, northwest territories)

nrcan_etl/src/energuide/embedded/region.py
buckley-w-david: Completely arbitrary trial and error with what seemed to give me good results.

sastels: sweet.

nrcan_etl/src/energuide/embedded/region.py
buckley-w-david: If the data is clean we should never get to this point, and it should exactly match either a province/territory name or its abbreviation.

I think this should be a warning since it's possible that this picks the wrong province, and it should be logged that we had to approximate it.

PR #457, opened 2018-04-03
Add instructions to add field to API

evadb:
exactly what is says on the tin

or is it

🤔

### DIFF
--- a/resource/add_to_api.md
+++ b/resource/add_to_api.md
@@ -1,5 +1,6 @@
 # Adding AIR50P and UGRAIR50PA (aka airLeakage) to the API
-This guide follows the instructions provided in the add_to_etl.md file. In order to add a field to the API, you will first
+
+This guide follows the instructions provided in the [add a field to ETL guide](https://github.com/cds-snc/nrcan_api/blob/master/resource/add_to_etl.md). In order to add a field to the API, you will first
 need to add it to the ETL. If you've already added your field to the ETL, you're good to go, read on!
 
 ## Loading your new data to your local mongoDB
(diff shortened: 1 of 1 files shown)


resource/add_to_api.md
```diff
@@ -0,0 +1,216 @@
...
+
+If you are adding a new `Dwelling` field, you would add it to the `Dwelling Type` test instead.
+
+2. In your console, run `yarn test`. All your tests should pass.
+
+## Adding an integration test for the new field
+
+### `api/test/queries.test.js`
+
+1. We need to test to make sure the field returns the data we expect. To do this, we need to pick a `houseID` and matching `airLeakage` values. You can get these values
```
buckley-w-david: `houseId` not `houseId`

resource/add_to_api.md
```diff
@@ -0,0 +1,216 @@
+# Adding AIR50P and UGRAIR50PA (aka airLeakage) to the API
+This guide follows the instructions provided in the add_to_etl.md file. In order to add a field to the API, you will first
```
buckley-w-david: It might be a good idea to link to the add_to_etl.md file

resource/add_to_api.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+Write down the first values for `houseId`, `measurement` & `upgrade` that the API returns. These will be your values for `testHouseId`, `testMeasurement` & `testUpgrade`
+in the following step.
+
+2. Above the `'retrieves all keys for ersRating data'` test, write the following test. Make sure to replace `testHouseId`, `testMeasurement` & `testUpgrade`
+with the values you wrote down in step 1.
+
+```
+it('retrieves all keys for airLeakage data', async () => {
```
buckley-w-david: I understand it, but it makes me vaguely uncomfortable that we're advising them to structure the tests to conform to the untested codes output

PR #464, opened 2018-04-05
Wall exception handling

buckley-w-david:
#450 was accidentally merged into the wrong base branch, this PR is a redo

### DIFF
--- a/nrcan_etl/tests/embedded/test_walls.py
+++ b/nrcan_etl/tests/embedded/test_walls.py
@@ -1,5 +1,8 @@
+import pytest
+
 from energuide.embedded import walls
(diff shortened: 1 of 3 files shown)


nrcan_etl/tests/embedded/test_walls.py
```diff
@@ -1,5 +1,8 @@
+import pytest
+
```
sastels: should this blank line be there?

buckley-w-david: I sometimes do that to separate the three sections of imports (standard, 3rd party, local package).

So yes intentional, but also fine with removing it

PR #466, opened 2018-04-06
Add houseType filter to API

evadb:
This PR adds houseType as a filter to the API. I think this is a cool & low effort filter to add, and will replace the furnace type in the poc 👍


buckley-w-david:
Good catch, I can clean it

Issue #448, opened 2018-03-28
Add documentation about default limits for API responses


pcraig3:
So we have not included a default limit in our API response, but we're using [`mongo-cursor-pagination`](https://www.npmjs.com/package/mongo-cursor-pagination) to return paginated data, and it's setting a default limit for us of 50 results. 
It's also setting a maximum number of responses (regardless of whatever is passed in) of 300.
We can override that ourselves, but we haven't done.

I read that here: https://www.npmjs.com/package/mongo-cursor-pagination#number-of-results

It's not documented in our API, but it should be. Otherwise this will give surprising results.

evadb:
This is now documented, closing 👍

pcraig3:
@obrien-j, Yeah, there's an environment variable we can set to override the default limit of 300, but we haven't done it.

I don't think there's a particular reason why not. We would probably want to add our own default if we did (otherwise, all queries with no explicit `limit` would return the entire dataset), but in principle I don't think there would be an issue.

evadb:
If we're discussing adding a new feature or modifying, maybe opening a new issue is best? This issue has been covered

pcraig3:
👍

PR #475, opened 2018-04-10
Add limit info to documentation

evadb:
This addresses this issue: https://github.com/cds-snc/nrcan_api/issues/448.

Adds documentation about how to use limits, unexpected limit results, & the default limit

### DIFF
--- a/docs/using_the_api.md
+++ b/docs/using_the_api.md
@@ -128,7 +128,7 @@ query {
 
 You should now be able to access any field you want from the API :tada:
 
-## Filters
+## Filters & limits
 
 The API includes filters to help narrow down data sets when querying for multiple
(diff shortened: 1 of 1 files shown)


docs/using_the_api.md
```diff
@@ -171,7 +171,27 @@ query {
...
+          upgrade
+        }
+      }
+    }
+  }
+}
+```
+
+**Important note:** The limit applies to the dwellings returned, not evaluations. Since a dwelling can have multiple evaluations, you will likely receive more
+than 30. The default limit set by the pagination is 50 dwellings.
```
buckley-w-david: We should also mention that regardless of what you say, there's a max limit of 300

evadb: 👍 good catch, just fixed that

